PR #5, opened 2020-03-06
RFC: White Flag

thibault-martinez:
[Rendered](https://github.com/thibault-martinez/protocol-rfcs/blob/rfc-white-flag/text/0005-white-flag/0005-white-flag.md)


text/0005-white-flag/0005-white-flag.md
```diff
@@ -0,0 +1,54 @@
...
+When a milestone is accepted by a node then the following algorithm to mutate the ledger state is performed:
+
+- Look for milestone whose index is the current latest solid milestone index + 1. Milestones must be applied in order!
+- When the milestone is found start traversing down the tangle in DFS going first to the trunk and only then the branch.
+Stop at the transactions that has only approved parents.
+- Start applying bundles to the diffmap only as you go up the recursion stack (when you climb up back from the leafs).
+- Every time you attempt to apply a bundle to the diffmap:
+If it is invalid, mark it as seen and ignored by the milestone.
+If it is valid and not conflicting with the current state then apply it to the ledger state. Mark it as seen and
+approved.
```
tsvisabo: approved or confirmed?
or maybe a different word because either one it eventually means that the mentioned bundle spent address (and spent into) are deducted/added in the ledger state

text/0005-white-flag/0005-white-flag.md
```diff
@@ -0,0 +1,54 @@
++ Feature name: `white-flag`
++ Start date: 2020-03-06
++ RFC PR: [iotaledger/protocol-rfcs#0005](https://github.com/iotaledger/protocol-rfcs/pull/5)
+
+# Summary
+
+Allow milestones to confirm conflicting bundles. Mutate the ledger state by deterministically ordering the bundles and
```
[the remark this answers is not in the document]
thibault-martinez: Added, thanks

text/0005-white-flag/0005-white-flag.md
```diff
@@ -0,0 +1,54 @@
...
+- Increase CTPS - Due to the above, increase in TPS and no left-behinds, we expect CTPS to increase as well.
+
+# Detailed design
+
+Let's define a conflicting bundle as a bundle that leads to a negative balance on an address if applied to the current
+ledger state.
+
+When a milestone is accepted by a node then the following algorithm to mutate the ledger state is performed:
+
+- Look for milestone whose index is the current latest solid milestone index + 1. Milestones must be applied in order!
```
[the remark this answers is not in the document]
thibault-martinez: I removed this part.

text/0005-white-flag/0005-white-flag.md
```diff
@@ -0,0 +1,103 @@
...
+Definitions:
+- confirm
+- ignore
+- reference / indirectly
+
+Let's define a conflicting bundle as a bundle that leads to a negative balance on an address if applied to the current
+ledger state.
+
+In case of conflicting bundles with White Flag, nodes will confirm only one bundle and ignore all the others. For this
+to work, they need to be sure they are all confirming the same bundle; hence, the need for a deterministic ordering
```
tsvisabo:
```suggestion
to work, nodes need to be sure they are all confirming the same bundle; hence, the need for a deterministic ordering
```

text/0005-white-flag/0005-white-flag.md
```diff
@@ -0,0 +1,103 @@
...
+of the tangle.
+
+## Deterministic ordering of the tangle
+
+<!-- TODO -->
+
+The deterministic ordering of the past-cone of a milestone is a [topological ordering](https://en.wikipedia.org/wiki/Topological_sorting)
+generated by a post-order Depth-First Search by going first through trunks, then branches and finally root.
+The stopping condition of this traversal is reaching bundles that are already confirmed.
```
tsvisabo: but what is the starting point of the DFS?
i guess the new milestone?
also, it's a bit confusing to say "The stopping condition of this traversal is reaching bundles that are already confirmed."
because all the transactions in this cone are confirmed, it's just that we stop if we reach transactions that were already confirmed by previous milestones

thibault-martinez: Right, already changed that in a future commit, thanks.

text/0005-white-flag/0005-white-flag.md
```diff
@@ -0,0 +1,155 @@
...
+        curr = s.peek()
+
+        // only apply if you reached a leaf or both approvees have been visited or confirmed by another milestone
+        if ((curr.trunk == null || visited.contains(curr.trunk) || curr.trunk.snapshot_index != milestone.index)
+          && ((curr.branch == null) || visited.contains(curr.branch) || curr.branch.snapshot_index != milestone.index)) {
+            ledger.apply(curr)
+            visited.add(curr)
+            s.pop()
+        }
+        else if (curr.trunk != null && !visited.contains(curr.trunk) && curr.trunk.snapshot_index == milestone.index) {
```
GalRogozinski: ditto

text/0005-white-flag/0005-white-flag.md
```diff
@@ -0,0 +1,155 @@
...
+            s.push(curr.trunk)
+        }
+        else if (curr.branch != null && !visited.contains(curr.branch) curr.branch.snapshot_index == milestone.index) {
```
GalRogozinski: ditto

text/0005-white-flag/0005-white-flag.md
```diff
@@ -0,0 +1,155 @@
...
+    s = new Stack()
+    visited = new Set()
+
+    s.push(milestone)
+
+    while (!s.empty()) {
+        curr = s.peek()
+
+        // only apply if you reached a leaf or both approvees have been visited or confirmed by another milestone
+        if ((curr.trunk == null || visited.contains(curr.trunk) || curr.trunk.snapshot_index != milestone.index)
```
GalRogozinski: using `milestone.index` s confusing for me, because I don't understand where it is updated...
I would rather use the pseudo-function `trunk.isMarkedByMilestone(milestone)` or `isConfirmed`...

GalRogozinski: I just realized that `curr.trunk == null` is a mistake

Because if it is null it may mean that we have a non-solid subtangle...
What the check here should be is if `curr.trunk == genesis`

Theoretically, `curr.trunk` should never be null because we solidify the milestone before applying it...
Maybe we should point this out in the RFC?

GalRogozinski: if it is null then there should be an error I believe, because we have holes in the tangle

thibault-martinez: Sure, I even wondered if it should be an SEP check instead of a genesis check but let's not over complicate it.

GalRogozinski: well yes actually
SEP is like genesis

text/0005-white-flag/0005-white-flag.md
```diff
@@ -0,0 +1,164 @@
...
+nodes allowing much easier tip selection algorithms leading to increased TPS;
+- By using this approach in combination with an appropriate TSA, during regular use, no honest transaction will ever
+require re-attaching leading to increased CTPS;
+- Doesn't come with added computation complexity by integrating nicely into already existing algorithms;
+
+# Detailed design
+
+First, let's define what it means for a bundle to be:
+
+- referenced (indirectly or not) by another bundle: it is in the past cone of this bundle;
```
Alex6323, line 28: When talking about relations between two or more bundles I would use labels A, B, ... to increase clarity.

text/0005-white-flag/0005-white-flag.md
```diff
@@ -0,0 +1,170 @@
...
+- confirmed: it is referenced by a milestone;
+- applied: it is confirmed and applied to the ledger state;
+- ignored: it is confirmed but not applied;
+- conflicting: it would lead to an invalid ledger state if applied;
+
+In case of conflicting bundles with White Flag, a node applies only one bundle to the ledger state and ignore all the
+others. For this to work, all the nodes need to be sure they are all applying the same bundle; hence, the need for a
+deterministic ordering of the Tangle.
+
+First, this RFC propose a deterministic ordering of the Tangle, then it explain which bundle is selected in case of
```
luca-moser, line 38:
```suggestion
First, this RFC proposes a deterministic ordering of the Tangle, then it explains which bundle is selected in case of
```

text/0005-white-flag/0005-white-flag.md
```diff
@@ -0,0 +1,170 @@
...
+conflict.
```
luca-moser, line 39:
```suggestion
conflicts.
```

text/0005-white-flag/0005-white-flag.md
```diff
@@ -0,0 +1,170 @@
...
+
+First, let's define what it means for a bundle to be:
+
+- referenced (indirectly or not) by another bundle: it is in the past cone of this bundle;
+- confirmed: it is referenced by a milestone;
+- applied: it is confirmed and applied to the ledger state;
+- ignored: it is confirmed but not applied;
+- conflicting: it would lead to an invalid ledger state if applied;
+
+In case of conflicting bundles with White Flag, a node applies only one bundle to the ledger state and ignore all the
```
luca-moser, line 34:
```suggestion
In case of conflicting bundles with White Flag, a node applies only one bundle to the ledger state and ignores all the
```

text/0005-white-flag/0005-white-flag.md
```diff
@@ -0,0 +1,170 @@
...
+
+# Detailed design
+
+First, let's define what it means for a bundle to be:
+
+- referenced (indirectly or not) by another bundle: it is in the past cone of this bundle;
+- confirmed: it is referenced by a milestone;
+- applied: it is confirmed and applied to the ledger state;
+- ignored: it is confirmed but not applied;
+- conflicting: it would lead to an invalid ledger state if applied;
```
Wollac, line 32:
```suggestion
First, let's define what it means for a bundle A to be:

- referenced (indirectly or directly) by bundle B: A is contained in the past cone of B;
- confirmed: A is referenced by a milestone;
- applied: A is confirmed and applied to the ledger state;
- ignored: A is confirmed but not applied;
- conflicting: A would lead to an invalid ledger state if applied;
```
I agree with @Alex6323, it think the above would be a tiny bit clearer.

text/0005-white-flag/0005-white-flag.md
```diff
@@ -0,0 +1,170 @@
...
+
+The feature presented in this RFC, White Flag, allows milestones to confirm conflicting bundles by enforcing
+deterministic ordering of the Tangle and applying only the first bundle(s) that does not violate the ledger state.
+
+The content of this RFC is based on [Conflict white flag: Mitigate conflict spamming by ignoring conflicts](https://iota.cafe/t/conflict-white-flag-mitigate-conflict-spamming-by-ignoring-conflicts/233).
+
+# Motivation
+
+- Eliminates the [Conflict spamming attack](https://iota.cafe/t/conflict-spamming-attack/232);
+- As conflicts are ignored in the balance computation, they don’t need to be considered during tip selection of the
```
Wollac, line 18:
```suggestion
- As conflicts are ignored in the balance computation, they do not need to be considered during tip selection of the
```

text/0005-white-flag/0005-white-flag.md
```diff
@@ -0,0 +1,170 @@
...
+nodes allowing much easier tip selection algorithms leading to increased TPS;
+- By using this approach in combination with an appropriate TSA, during regular use, no honest transaction will ever
+require re-attaching leading to increased CTPS;
+- Doesn't come with added computation complexity by integrating nicely into already existing algorithms;
```
Wollac, line 22:
```suggestion
- Does not come with added computation complexity by integrating nicely into already existing algorithms;
```

text/0005-white-flag/0005-white-flag.md
```diff
@@ -0,0 +1,170 @@
...
+others. For this to work, all the nodes need to be sure they are all applying the same bundle; hence, the need for a
+deterministic ordering of the Tangle.
+
+First, this RFC proposes a deterministic ordering of the Tangle, then it explains which bundle is selected in case of
+conflicts.
+
+**Note: this RFC is about ledger computation only. For this reason, it assumes that the past cone of a milestone has
+already been confirmed and all the referenced bundles have been validated. However, in the event that an invalid bundle
+was encountered in the confirmation traversal, a node's expected behaviour would be to completely stop operations - and
+log the error - as it would mean the coordinator confirmed something it shouldn't have.**
```
Wollac, line 44:
```suggestion
log the error - as it would mean the coordinator confirmed something it should not have.**
```

text/0005-white-flag/0005-white-flag.md
```diff
@@ -0,0 +1,170 @@
...
+starting from a milestone and by going first through `trunk` bundle, then `branch` bundle and finally current bundles.
+Since only a subset of bundles is considered, the stopping condition of this DFS is reaching bundles that are already
+confirmed by another milestone.
+
+## Applying first bundle(s) that does not violate the ledger state
+
+If a conflict is occurring in the set of bundles confirmed by a milestone, nodes have to apply the first - with regards
+to the order previously proposed - of the conflicting bundles to the ledger and ignore all the others.
+
+Once a bundle is marked as ignored, this is final and can't be changed by a later milestone.
```
Wollac, line 70:
```suggestion
Once a bundle is marked as ignored, this is final and cannot be changed by a later milestone.
```

text/0005-white-flag/0005-white-flag.md
```diff
@@ -0,0 +1,170 @@
...
+        }
+    }
+}
+```
+
+**Notes**:
+- Even though the tangle is a graph made of transactions, for the sake of this pseudo-code it is considered as a graph
+of bundles. For this reason, when the `trunk` and/or the `branch` of a bundle are mentioned, they are actually referring
+to the `trunk` and/or `branch` of the last transaction of the bundle.
+- `solid_entry_points` is a set of hashes that are considered solid even though we don't have them or their past in
```
Wollac, line 118:
```suggestion
- `solid_entry_points` is a set of hashes that are considered solid even though we do not have them or their past in
```

text/0005-white-flag/0005-white-flag.md
```diff
@@ -0,0 +1,170 @@
...
+- Eliminates the [Conflict spamming attack](https://iota.cafe/t/conflict-spamming-attack/232);
+- As conflicts are ignored in the balance computation, they don’t need to be considered during tip selection of the
+nodes allowing much easier tip selection algorithms leading to increased TPS;
+- By using this approach in combination with an appropriate TSA, during regular use, no honest transaction will ever
+require re-attaching leading to increased CTPS;
+- Doesn't come with added computation complexity by integrating nicely into already existing algorithms;
+
+# Detailed design
+
+First, let's define what it means for a bundle to be:
```
Wollac, line 26:
```suggestion
First, let us define what it means for a bundle to be:
```

jakubcech:
Snapshot signing required for White flag deployment will be a separate RFC.

PR #8, opened 2020-03-10
RFC: Weighted Uniform Random Tip-Selection

luca-moser:
[Rendered](https://github.com/luca-moser/protocol-rfcs/blob/rfc-urts-tip-sel/text/0008-weighted-uniform-random-tip-selection/0008-weighted-uniform-random-tip-selection.md)


gjeee:
The concept (to me at least) of "confirmed root transactions" is a bit unclear (yellow boxes in the pic). But what are the green ones then? I would think they are as well. 
image 

Here again...what are the green ones then? 
image

gjeee:
Also I think TSI is not exactly defined:

> Transaction Snapshot Index (TSI) defines the index of the milestone which confirmed a given transaction.

Shouldn't it be index of the oldest milestone?

luca-moser:
@gjeee First thanks for your comments, please use the review feature next time as the way you wrote your comments they don't target specific lines (even though they should/do)

Regarding your first comment:
> The concept (to me at least) of "confirmed root transactions" is a bit unclear (yellow boxes in the pic). But what are the green ones then? I would think they are as well.

The text from the RFC says:
> Confirmed Root Transactions defines the set of first seen transactions which are confirmed by a previous milestone when we walk the past cone of a given transaction. The walk stops on confirmed transactions.

image

The green transactions are not `Confirmed Root Transactions` because if you would walk as described in the text from the PoV transaction, you wouldn't reach them as you stop walking as soon as you reach an already confirmed `Confirmed Root Transaction`. Those "stopping points" are the yellow transactions as described in the RFC.

Regarding your second comment:
> Also I think TSI is not exactly defined:
>>Transaction Snapshot Index (TSI) defines the index of the milestone which confirmed a given transaction.

> Shouldn't it be: ......the index of the oldest milestone......?

Only ever one milestone does confirm a given transaction, there's no need to say oldest in this context. Of course newer milestone still reference previous transactions but in this context it is clear that we're talking about the milestone which confirms a transaction.

text/0008-weighted-uniform-random-tip-selection/0008-weighted-uniform-random-tip-selection.md
```diff
@@ -0,0 +1,190 @@
...
+
+Definitions:
+* `Direct Approvers` are the set of transactions which directly approve a given transaction.
+* `Approvee` is the directly approved transaction of a given transaction.
+Example: the trunk/branch transactions are the approvees of a given transaction.
+* `Solid` means that the past cone of a given transaction exists in the database.
+* A `tail transaction` is the transaction at index zero of a bundle. Only a tail transaction can be a `tip`.
+* A `tip` is a solid tail transaction of a bundle without any approvers. Its past cone contains only structural valid bundles.
+* A `score` is a scoring  determining the likeliness to select a given `tip`.
+* `Confirmed Root Transactions` defines the set of first seen transactions which are confirmed by a previous milestone
```
Wollac, line 31: It should be made clear that the milestone itself is also a "Confirmed Root Transactions" since it effectively confirms itself. Otherwise the linear tangle, i.e. branch always equal to trunk, would not have any Confirmed Root Transactions, rendering the tip score computation impossible.

text/0008-weighted-uniform-random-tip-selection/0008-weighted-uniform-random-tip-selection.md
```diff
@@ -0,0 +1,190 @@
...
+* `Youngest Transaction Root Snapshot Index (YTRSI)` defines the highest milestone index of a set of
+`Confirmed Root Transactions` of a given transaction.
+* `Latest Solid Milestone Index (LSMI)` the index of the latest solid milestone.
+* `Below Max Depth (BMD)` defines a threshold value up on which it is decided on whether a transaction is not
+relevant in relation to the recent parts of the Tangle. The current `BMD` for mainnet nodes is 15 milestones, 
+which means that transactions of which their `OTRSI` is more than 15, are "below max depth".
+
+### OTRSI / YTRSI example
+Given the blue PoV transaction, the `OTRSI` of it is milestone 1 and `YTRSI` milestone 2. The orange
+and purple transactions are the `Confirmed Root Transactions`.
```
Wollac, line 47: The red milestones are also `Confirmed Root Transactions`

text/0008-weighted-uniform-random-tip-selection/0008-weighted-uniform-random-tip-selection.md
```diff
@@ -0,0 +1,190 @@
...
+```
+
+var tips = Set(tips_and_score)
+
+func select() Tip {
+    // compute the sum of the score of all tips
+    scoreSum := tips.ScoreSum()
+    
+    // get a random number between 1 and the score sum
+    r := rand.Int(1, scoreSum)
```
Wollac, line 143: This will fail, if there are only lazy tips as then `scoreSum == 0`. What should we do in that particular case?

Although this is a edge-case, it could still happen when the network connections are lost and only the coordinator was able to issue milestones in the recent past.

text/0008-weighted-uniform-random-tip-selection/0008-weighted-uniform-random-tip-selection.md
```diff
@@ -0,0 +1,190 @@
...
+
+# Motivation
+
+Because of the `white-flag` confirmation algorithm, it is no longer necessary to perform complex
+tip-selection which evaluates ledger mutations while walking, therefore, a simpler and better 
+performing algorithm can be used to select tips, which in turn increases overall transaction throughput.
+
+In order to maximize confirmation rate however, the algorithm needs to return tips which are `non-lazy`.
+Non-lazy in this context means, that a tip does not attach to a cone of transactions which is too far
+in the past, as such cone is likely to be already confirmed and therefore doesn't contribute to the
```
Wollac, line 18:
```suggestion
in the past, as such cone is likely to be already confirmed and therefore does not contribute to the
```
avoid contractions

text/0008-weighted-uniform-random-tip-selection/0008-weighted-uniform-random-tip-selection.md
```diff
@@ -0,0 +1,190 @@
...
+
+# Unresolved questions
+
+#### When to compute the `YTRSI`/`OTRSI` of a transaction?
+It is not yet clear when or how often the `YTRSI`/`OTRSI` values of a transaction should be updated.
+If the values are only computed once after a transaction became solid, the `YTRSI`/`OTRSI` might not
+resemble the true values, as subsequent milestones might confirm transactions within the same cone the
+given transaction approved.
+
+However, one can argue that the `YTRSI`/`OTRSI` don't shift by that much in
```
Wollac, line 187:
```suggestion
However, one can argue that the `YTRSI`/`OTRSI` do not shift by that much in
```

text/0008-weighted-uniform-random-tip-selection/0008-weighted-uniform-random-tip-selection.md
```diff
@@ -0,0 +1,190 @@
...
+such case, so that it is fine to compute them only up on solidification of a given transaction.
+This assumption builds up on the fact, that the Coordinator wouldn't confirm transactions
```
Wollac, line 189:
```suggestion
This assumption builds up on the fact, that the Coordinator would not confirm transactions
```

text/0008-weighted-uniform-random-tip-selection/0008-weighted-uniform-random-tip-selection.md
luca-moser: If there are only lazy tips, then `tips` is empty and we return null (I've added a corresponding code snippet to catch that condition.). If the node is not synchronized or has no network connection, it should not execute tip-selection.

text/0008-weighted-uniform-random-tip-selection/0008-weighted-uniform-random-tip-selection.md
luca-moser: Added a mention.

text/0008-weighted-uniform-random-tip-selection/0008-weighted-uniform-random-tip-selection.md
luca-moser: Added a note.

text/0008-weighted-uniform-random-tip-selection/0008-weighted-uniform-random-tip-selection.md
Wollac: I agree. Since it has now made clear that `Confirmed Root Transactions` also contain milestones, we will always at least have the milestone as a non-lazy tip for a synched node

muXxer:
A tip should stay in the tip pool for some time, even if it got referenced by other transactions.
Otherwise the tangle would become really slim and it would be much harder to sync, because we have to request tx by tx.

Maybe have two criteria at the same time? A minimum time after first reference by another tx or a maximum amount of references. If one of the two criteria is reached, the tip is removed from the pool.

luca-moser:
I've added a paragraph for those rules, please verify whether they are ok @muXxer.

GalRogozinski:
2 points we need to agree on:
1. Only bundles with valid signatures should be legible for tip selection
> currently we think yes
2. Do we need to add a delay before tip become legible for tip selection. Maybe delay from (1) is enough?

text/0008-weighted-uniform-random-tip-selection/0008-weighted-uniform-random-tip-selection.md
```diff
@@ -0,0 +1,205 @@
...
+![sdf](images/cnf_tx_roots.PNG)
+* `Transaction Snapshot Index (TSI)` defines the index of the milestone which confirmed a given transaction.
+* `Oldest Transaction Root Snapshot Index (OTRSI)` defines the lowest milestone index of a set of
+`Confirmed Root Transactions` of a given transaction.
+* `Youngest Transaction Root Snapshot Index (YTRSI)` defines the highest milestone index of a set of
+`Confirmed Root Transactions` of a given transaction.
+* `Latest Solid Milestone Index (LSMI)` the index of the latest solid milestone.
+* `Below Max Depth (BMD)` defines a threshold value up on which it is decided on whether a transaction is not
+relevant in relation to the recent parts of the Tangle. The current `BMD` for mainnet nodes is 15 milestones, 
+which means that transactions of which their `OTRSI` is more than 15, are "below max depth".
```
thibault-martinez, line 45: > of which their `OTRSI` is more than 15, are "below max depth".

Shouldn't it be in relation to LSMI instead of just OTRSI ?

text/0008-weighted-uniform-random-tip-selection/0008-weighted-uniform-random-tip-selection.md
```diff
@@ -0,0 +1,205 @@
...
+    approveesOTRSICheck := 2
+    for (i := 0; i < 2; i++) {
+        approvee := approvees[i]
+    
+        // direct approvee is already lazy, therefore so is this tip
+        if (approvee.Score == 0) {
+            return Score.LAZY
+        }
+        
+        // if the OTRSI to LSMI delta of the approvee is C2, we mark it as such
```
thibault-martinez, line 111: A missing word maybe ? `is C2` ?

text/0008-weighted-uniform-random-tip-selection/0008-weighted-uniform-random-tip-selection.md
luca-moser: Adjusted.

text/0008-weighted-uniform-random-tip-selection/0008-weighted-uniform-random-tip-selection.md
luca-moser: `C2` is defined above as ` = 7 milestones`.

luca-moser: Ah, I see what you meant. Wrote "above" in front of it.

PR #9, opened 2020-03-10
RFC: Ed25519 signature scheme

Wollac:
[Rendered](https://github.com/Wollac/protocol-rfcs/blob/ed25519/text/0009-ed25519-signature-scheme/0009-ed25519-signature-scheme.md)

Example Go implementation: [wollac/iota-crypto-demo](https://github.com/Wollac/iota-crypto-demo)


text/0009-ed25519-signature-scheme/0009-ed25519-signature-scheme.md
```diff
@@ -0,0 +1,112 @@
...
+
+# Rationale and alternatives
+
+- The Ed25519 public-key signature system is a standard for modern signature schemes, as it offers high-speed high-security signatures.
+- This RFC aims for a seamless integration of Ed25519 signatures in the existing protocol. As such, the transaction structure as well as the resulting address format remain unchanged. While this offers the least possible development overhead to adapt existing W-OTS-only implementations, it also comes with a few drawbacks:
+  - W-OTS and Ed25519 addresses look exactly the same. This makes it impossible for the user or the system to apply any special considerations for the different signature schemes based on the addresses alone.
```
rajivshah3, line 97: Will Ed25519 addresses also have checksums when shown to the user? If so, systems could figure out whether the address is W-OTS or Ed25519 by recalculating the checksum (assuming Ed25519 address checksums are calculated differently from W-OTS address checksums)

text/0009-ed25519-signature-scheme/0009-ed25519-signature-scheme.md
```diff
@@ -0,0 +1,112 @@
...
+  - Signing the bundle hash with Ed25519 leads to unnecessary double hashing of the bundle essence. The performance of the signature scheme could be improved further by signing the bundle essence directly with Ed25519.
+  - One of the compelling features of Ed25519 is that it provides very small binary key and signature sizes. By representing these as fixed length ternary strings, a lot of this potential is lost with respect to the transaction size. However, it is important to note that even simple compression algorithms, like run-length encoding, mostly circumvent this issue. 
+- Even though W-OTS offers strong quantum robustness, it comes with huge drawbacks for a payment system. Most notably here is the fact that signing more than once with the same key endangers the funds. A malicious wallet, a single malicious node or even network issues inducing wrongly reported balances, can force an honest user to sign one or more invalid transactions. Although these transactions will never be accepted by the network, they still expose valid signatures which can be used to break the private key even on conventional today's hardware.
```
rajivshah3, line 101:
```suggestion
- Even though W-OTS offers strong quantum robustness, it comes with huge drawbacks for a payment system. Most notably here is the fact that signing more than once with the same key endangers the funds. A malicious wallet, a single malicious node or even network issues inducing wrongly reported balances, can force an honest user to sign one or more invalid transactions. Although these transactions will never be accepted by the network, they still expose valid signatures which can be used to break the private key even on today's conventional hardware.
```

Tiny grammar fix (`conventional today's` -> `today's conventional`)

text/0009-ed25519-signature-scheme/0009-ed25519-signature-scheme.md
Wollac: It was the goal of this RFC to not change anything in the current IOTA protocol transaction layout. As such, even Ed25519 addresses would need to be represented in 243 trits.
Unfortunately, due to the possibility of replay-attacks this entire idea is not feasible and Ed25519 needs to be combined with transaction layout changes, where we can indeed have different address versions

Wollac:
Ed25519 will be introduced together with the UTXO model to offer replay protection.
As such, an RFC trying to 100% maintain the legacy transaction layout no longer makes sense.

PR #11, opened 2020-05-06
RFC: UTXO Model

hmoog:
[Rendered Version](https://github.com/hmoog/protocol-rfcs/blob/master/text/0011-utxo-model/0011-utxo-model.md)


text/0011-utxo-model/0011-utxo-model.md
```diff
@@ -0,0 +1,194 @@
...
+to combine two eventually conflicting sub-tangles. 
+
+The UTXO model uses a different form of record keeping which enables the validation of transactions in constant time
+`O(1)` and vastly improves the "expressiveness" of the ledger state by enabling things like:
+
++ **Colored Coins:** IOTA tokens can be marked with a certain "color". This color is retained throughout transfers and
+gives the tokens a certain "meaning" (i.e. tokenized assets, resource- and access-tokens ...).
+  
++ **Scalable Layer1 Smart Contracts:** Balances can be extended by a sequence of dynamic unlock conditions that will
+enable the use of a non-touring complete scripting language for smart contracts on layer1, which would not just be run
```
laumair, line 29: non-turing

laumair, line 29: Typo: `non-turing`

hmoog: Thx :P

hmoog: Thx

text/0011-utxo-model/0011-utxo-model.md
```diff
@@ -0,0 +1,194 @@
...
+
+Even though Bitcoin has this kind of inherent agnosticism towards using addresses, it has proven to be extremely
+beneficial to at least have some form of *logical identifier* that groups related outputs together and that can also be
+used to *authorize* payments in the form of address signatures. It is for that very reason that the predominant part of
+transaction executed on today's DLTs make use of addresses.
+
+To build a bridge between the worlds of *UTXO-* and *account-based* ledgers, we define our UTXO variant to use
+addresses instead of output indexes as this additional identifier:
+
+``OutputID = Pair<Address, TransactionID>``
```
GalRogozinski, line 80: Do you want to have 2 indexes for the output to be used internally by the node when it tracks utxos?
Or do you want this to be the actual id to be used inside the transaction that we create?

I understood that you want the latter. If so, I just don't understand the motivation. I think it just bloats up the transaction. If I understood correctly, then could you please elaborate on why it is useful?

text/0011-utxo-model/0011-utxo-model.md
```diff
@@ -0,0 +1,194 @@
...
+
+All **colored balances** sent to the same address (by the same transaction) becomes part of the same output.
```
GalRogozinski, line 82: Clarification:
Colored balances include or don't include the base color?

I assume that no since you highlighted this, but then why should they behave differently?

text/0011-utxo-model/0011-utxo-model.md
```diff
@@ -0,0 +1,194 @@
...
+transferred:
+
+``Color = Array<byte>``
+
+Next to *user defined* colors, there are two *builtin color* values that carry a special meaning:
+
+- ``COLOR_IOTA = Array<byte>(0, 0, 0, ..., 0)`` represents the *base color* of uncolored coins.
+
+- ``COLOR_MINT = Array<byte>(255, 255, 255, ..., 255)`` represents a color which is being replaced by the
+  ``Transaction ID`` before being booked as an output. It is consequently used to "mint" new colored coins.
```
GalRogozinski, line 100: Transaction ID will become a color?

I think I need more clarifications here :sweat_smile: 
Didn't really understand this line

text/0011-utxo-model/0011-utxo-model.md
```diff
@@ -0,0 +1,194 @@
...
+a marshaled version of these parameters.
+
+``OPCodeMetadata = Array<byte>`` 
+
+As long as these checks are not much more *expensive* than signature checks, it does not introduce problems but
+massively extends the functionality of the protocol. We have to however be very careful with adding new conditions, to
+not accidentally introduce potential bottlenecks.
+
+In contrast to bitcoin, we do not add a fully scriptable outputs (because that could result in long execution times),
+but we offer predefined methods for different use cases.
```
GalRogozinski, line 138: What I understood is that by using `OP` + `OPCodeMetadata` instead of a script, you limit one op code use per output. We also don't have a stack.

Did I understand correctly?

About the long execution times, I guess that if we will use a script instead we can simply have a max length limit on it.
This would solve the problem no?

Even if it does, I don't know yet whether I am for scripts or not..
However, every time we would want a new functionality, people running old node version won't support it.
With scripting it will be more versatile, so we need to take this into account

text/0011-utxo-model/0011-utxo-model.md
```diff
@@ -0,0 +1,194 @@
...
+
+For now, we simply reserve the additional fields in the output but hard code the OPCode to be ``SIG_CHECK = OPCode(0)``.
+
+### Conflict Detection
+
+The biggest benefits of this ledger state is the fast and easy detection of double spends and invalid transaction:
+
+- A transaction is valid, if the sum of the moved funds is ``0`` and it only spends colors in the amounts that were
+available in the used inputs. It is however possible to override the color of the consumed tokens using either the
+``MINT_COLOR`` color to create a new colored token or by using the ``IOTA_COLOR`` to remove the color.
```
GalRogozinski, line 148: I really don't understand the whole IOTA colored coins concept

So if I color a coin and then transfer it to you, then you could change the color or even remove it?
Don't understand the use-case here

text/0011-utxo-model/0011-utxo-model.md
```diff
@@ -0,0 +1,194 @@
...
+
+  Instead of having to walk through multiple transaction to eventually find the spent balance in the tangle, we can now
+instantly load the corresponding funds by accessing the consumed outputs. A transaction that tries to spend non-existing
+funds would never become solid.
```
GalRogozinski, line 152: you mean `valid` no?

Or has the definition for `solid` change to include validity and I didn't know about it :-P ?

text/0011-utxo-model/0011-utxo-model.md
```diff
@@ -0,0 +1,194 @@
...
+the additional identifier for outputs, allows us to bundle multiple colored balances in a single output, which leads to
+a smaller fragmentation of the funds as if we would give every color a separate output.
+
+If we decide to not go with this approach, then we will essentially limit ourselves to just a few hundred TPS because of
+the massive overhead of checking incoming transactions. In addition, the conflict detection will be much harder.
+
+# Unresolved questions
+
+- We need to think about good unlock opcodes and options.
+- We should define a PR that specifies the "branch based ledger state" as this is very closely related to the
```
GalRogozinski, line 193: a few words or a link on "branch based ledger state"?

text/0011-utxo-model/0011-utxo-model.md
hmoog: This has nothing to do with validity.

If I create a transaction that references an output `Pair<AddressX, TransactionQ>` but I haven't received `TransactionQ`, yet then this means that the outputs of `TransactionQ` are also not existing and it is not solid.

If I receive the `TransactionQ` later on then it becomes solid and can be booked. If it is however referencing an output that does not exist, then it will simply stay unsolid until it gets cleaned up.

jakubcech:
https://hackmd.io/@uwFzh-3_QZOoeCuKDppKnQ/HyM3P8WoL

text/0011-utxo-model/0011-utxo-model.md
```diff
@@ -0,0 +1,194 @@
...
+In contrast to bitcoin, we do not add a fully scriptable outputs (because that could result in long execution times),
+but we offer predefined methods for different use cases.
+
+For now, we simply reserve the additional fields in the output but hard code the OPCode to be ``SIG_CHECK = OPCode(0)``.
+
+### Conflict Detection
+
+The biggest benefits of this ledger state is the fast and easy detection of double spends and invalid transaction:
+
+- A transaction is valid, if the sum of the moved funds is ``0`` and it only spends colors in the amounts that were
```
Wollac, line 146: There is more to validity:
- There must be at least one output (and at least one input)
- the output referenced by each OutputID in the input list must be valid and exists.
- the total output sum (per color) must be equal to the sum (per color) over all outputs
- only one output per address
- What about 0-value inputs/outputs are they valid?
- ...

text/0011-utxo-model/0011-utxo-model.md
Wollac: I really think a separate RFC would help here. There we should also motivate the whole idea behind this particular concept.

text/0011-utxo-model/0011-utxo-model.md
Wollac, line 152: I think the concept of "Solidity" in the context of UTXO is not that simple and it must be explained in more detail.
This is also the first time "solid" is mentioned at all without any definition.

text/0011-utxo-model/0011-utxo-model.md
```diff
@@ -0,0 +1,194 @@
...
+- Accordingly, a transaction is performing a double spend when there is already another transaction consuming the same
+output.
+
+# Drawbacks
+
+- A UTXO-based ledger is much more complex in the implementation than just a dictionary of balances.
+- A UTXO-based ledger stores more information and therefore requires "potentially" more disk space.
+  
+  However, if
+sweeping becomes a central part of the protocol with all wallets supporting it, then this solution will most
```
Wollac, line 163: What is "sweeping"? Either define it or remove this reference

text/0011-utxo-model/0011-utxo-model.md
```diff
@@ -0,0 +1,194 @@
...
+probably not perform much worse than an account-based ledger.
+
+  Considering how we store ledger state today, with several *checkpoints* where we *cache the ledger state* to reduce
+  possible tangle walking time, it might also be possible that the resource consumption drops because we are no longer
+  required to have these additional checkpoints with their corresponding ledger state (in memory and on disk).
+
+  In general, it does not introduce a new attack vector because it is already today possible to spread out small
```
Wollac, line 170: Maybe mention the name "dust attack" here?

text/0011-utxo-model/0011-utxo-model.md
```diff
@@ -0,0 +1,194 @@
...
+instantly load the corresponding funds by accessing the consumed outputs. A transaction that tries to spend non-existing
+funds would never become solid.
+
+- Accordingly, a transaction is performing a double spend when there is already another transaction consuming the same
+output.
+
+# Drawbacks
+
+- A UTXO-based ledger is much more complex in the implementation than just a dictionary of balances.
+- A UTXO-based ledger stores more information and therefore requires "potentially" more disk space.
```
Wollac, line 160: In this section we go from UTXO might require more space to UTXO might require less space. Can we be more precise here? Under which exact conditions does it require more under which less? If we are unsure about this consider remove this point completely.

text/0011-utxo-model/0011-utxo-model.md
```diff
@@ -0,0 +1,194 @@
...
+``MINT_COLOR`` color to create a new colored token or by using the ``IOTA_COLOR`` to remove the color.
+
+  Instead of having to walk through multiple transaction to eventually find the spent balance in the tangle, we can now
+instantly load the corresponding funds by accessing the consumed outputs. A transaction that tries to spend non-existing
+funds would never become solid.
+
+- Accordingly, a transaction is performing a double spend when there is already another transaction consuming the same
+output.
+
+# Drawbacks
```
Wollac, line 157: Should we maybe refer to Ethereums rationale that UTXO introduces "states" even for one address/account making it more complicated in smart contracts?

Wollac, line 157: Should we mention "fungibility" as a drawback, since with UTXO the origin of every coin can be derived?

text/0011-utxo-model/0011-utxo-model.md
Wollac, line 160: Transaction will definitely get larger: everything that can expressed with one input in the account model needs at least one input in UTXO. Thus, UTXO tx size ≥ account tx size.

text/0011-utxo-model/0011-utxo-model.md
Wollac: I think that concepts like reality/branch based ledgers is many layers away from plain UTXO transactions and relies on much more concepts. So, this should not be mentioned as an unresolved question for plain UTXO at all.

text/0011-utxo-model/0011-utxo-model.md
dysbulic: I stumbled over "`solid`" as well as an unfamiliar terms.

Maybe add a note or footnote or something.

Wollac:
superseded by https://github.com/iotaledger/protocol-rfcs/pull/18

PR #15, opened 2020-06-10
RFC: Binary-to-Ternary Encoding

Wollac:
[Rendered](https://github.com/Wollac/protocol-rfcs/blob/binary-to-ternary/text/0015-binary-to-ternary-encoding/0015-binary-to-ternary-encoding.md)

Example Go implementation in [wollac/iota-crypto-demo](https://github.com/Wollac/iota-crypto-demo):
- Binary-to-Ternary Encoding: [pkg/encoding/ternary/ternary.go](https://github.com/Wollac/iota-crypto-demo/blob/master/pkg/encoding/ternary/ternary.go)

### DIFF
--- a/text/0015-binary-to-ternary-encoding/0015-binary-to-ternary-encoding.md
+++ b/text/0015-binary-to-ternary-encoding/0015-binary-to-ternary-encoding.md
@@ -53,7 +53,8 @@ foreach 6-trit group g in T:
 
 # Drawbacks
 
-Conceptually, one byte can be encoded using log<sub>3</sub>(256) ≈ 5.0474 trits. Thus, encoding 1 byte as 6 trits consumes considerably more memory than the mathematical minimum.
+- Conceptually, one byte can be encoded using log<sub>3</sub>(256) ≈ 5.0474 trits. Thus, encoding 1 byte as 6 trits consumes considerably more memory than the mathematical minimum.
+- Depending on the actual implementation the conversion might be malleable: E.g. both `Z9` (-1) and `LI`(255) could be decoded as `ff`. However, `LI` can never be the result of a valid encoding. As such, the implementation must reject such invalid inputs.
 
 # Rationale and alternatives
 
--- a/text/0015-binary-to-ternary-encoding/0015-binary-to-ternary-encoding.md
+++ b/text/0015-binary-to-ternary-encoding/0015-binary-to-ternary-encoding.md
@@ -64,11 +64,7 @@ There are several ways to convert binary data into ternary, e.g.
  - the conversion used as part of the [Kerl](https://github.com/iotaledger/kerl/blob/master/IOTA-Kerl-spec.md) hash function encoding chunks of 48 bytes as 242 trits,
  - or by encoding each bit as one trit with the corresponding value.
 
-Each conversion method has different advantages and disadvantages. However, since the `t5b1` encoding is well-defined and has been used in [IRI](https://github.com/iotaledger/iri) for both network communications and storage layers for a long time, choosing the direct counterpart for the opposite conversion represents the most logical solution providing a nice balance between performance and memory-efficiency.
-
-# Open questions
-
-The current client libraries do not offer any functionality to convert bytes into trits. The closest offered functionality is the ASCII to trit conversion, which is used for human-readable messages in transactions:
+The current client libraries do not provide any functionality to convert an arbitrary amount of bytes into trits. The closest available functionality is the ASCII to trit conversion, which is used for human-readable messages in transactions:
 ```
 T ← []
 foreach char c in S:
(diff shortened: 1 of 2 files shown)


text/0015-binary-to-ternary-encoding/0015-binary-to-ternary-encoding.md
```diff
@@ -0,0 +1,79 @@
...
+  - ternary (trytes): `GWLW9DLDDCLAJDQXBWUZYZODBYPBJCQ9NCQYT9IYMBMWNASBEDTZOYCYUBGDM9C9`
+
+# Drawbacks
+
+- Conceptually, one byte can be encoded using log<sub>3</sub>(256) ≈ 5.0474 trits. Thus, encoding 1 byte as 6 trits consumes considerably more memory than the mathematical minimum.
+- Depending on the actual implementation the conversion might be malleable: E.g. both `Z9` (-1) and `LI`(255) could be decoded as `ff`. However, `LI` can never be the result of a valid encoding. As such, the implementation must reject such invalid inputs.
+
+# Rationale and alternatives
+
+There are several ways to convert binary data into ternaray, e.g.
```
luca-moser, line 61:
```suggestion
There are several ways to convert binary data into ternary, e.g.
```

text/0015-binary-to-ternary-encoding/0015-binary-to-ternary-encoding.md
```diff
@@ -0,0 +1,79 @@
...
++ RFC PR: [iotaledger/protocol-rfcs#0015](https://github.com/iotaledger/protocol-rfcs/pull/15)
+
+# Summary
+
+In the IOTA protocol, a transaction is represented as ternary data. However, sometimes it is necessary to store binary data (e.g. the digest of a binary hash function) inside of a transaction. This requires the conversion of binary into ternary strings.
+The IOTA client libraries support the opposite conversion that encodes 5 trits as 1 byte (sometimes also referred to as `t5b1`), which is used for network communication and in storage layers. This RFC describes the corresponding counterpart to encode 1 byte as 6 trits.
+
+# Motivation
+
+A byte is composed of 8 bits that can represent 2<sup>8</sup> = 256 different values. On the other hand, 6 trits can hold 3<sup>6</sup> = 729 values while 5 trits can hold 3<sup>5</sup> = 243 values. Therefore, the most memory-efficient way to encode one byte requires the use of 6 trits. Although there exist many potential encoding schemes to convert binary data into ternary, the proposed version has been designed to directly match the widely used `t5b1` encoding.
```
GalRogozinski, line 12: > Therefore, the most memory-efficient way to encode one byte requires the use of 6 trits.

1. That sentence is very true for one byte. But when I read it I paused and I started to think about what will happen if we want to convert more than one byte. For example if we try to take the number of bytes that make up a hash there is a more efficient encoding. Maybe we should mention it so future readers won't pause and ask questions?

2. Since we do `t5b1` it makes sense that we can do `b1t5`. In IRI there is a `Converter#getTrits` that does so using an internal mapping. I think (not sure) that bee has (or used to have) something like this as well.
I recall that somewhere you had discussions with @thibault-martinez about this. An explanation on the RFC would also be welcome :-)

text/0015-binary-to-ternary-encoding/0015-binary-to-ternary-encoding.md
```diff
@@ -0,0 +1,79 @@
...
+
+### Bytes to trits
+In order to encode a binary string S into ternary, each byte of S is interpreted as a signed (two's complement) 8-bit integer value v. Then, v is encoded as a little-endian 6-trit string in balanced ternary representation. Finally, the resulting groups of trits are concatenated.
+
+This algorithm can also be described using the following pseudocode:
+```
+T ← []
+foreach byte b in S:
+  v ← int8(b)
+  g ← IntToTrits(v, 6)
```
GalRogozinski, line 24: So `IntToTrits` ( and`TritsToInt`) is a blackbox?

Just from reading the RFC I wouldn't be able to implement it...
Maybe you want the developer to depend on you examples?
If so can you make it clearer please?

text/0015-binary-to-ternary-encoding/0015-binary-to-ternary-encoding.md
GalRogozinski: I think a bit more down in rational and alternatives you addressed (1)

alexsporn:
As discussed with @Wollac I think we need both t5b1/1b5t and t6b1/1b6t for the current nodes, since we are already using 1b5t to decode the trytes from their storage (in bytes).
Regarding the implementation it should probably be different encoding packages. I think Bee is already doing this.

text/0015-binary-to-ternary-encoding/0015-binary-to-ternary-encoding.md
Wollac: It is out of the scope of the protocol RFCs to formally define aspects of the current protocol. For this, the reader needs to refer to existing docs/spec outside of the repository.

Wollac:
I've added clarifications to address @alexsporn and @GalRogozinski remarks.

text/0015-binary-to-ternary-encoding/0015-binary-to-ternary-encoding.md
Wollac: Made the difference between `t1b5` and `b1t6` clearer in an additional paragraph.
I hop that addresses this comment.

PR #16, opened 2020-07-13
RFCs mdbook

thibault-martinez:


nothingismagick:
Wish this would make a PDF too.

Wollac:
Are there configuration options to closer match the IOTA design styles?

thibault-martinez:
Are you speaking about `no-section-label` ? By default mdbook creates sections labels (1.1, 1.2, ...) but it makes no sense for RFCs.

Wollac:
No, I was more thinking about adding the IOTA logo or using the color palette to make it closer to existing iota.org websites.

thibault-martinez:
Ahhh, I'm sure we can change the theme yes.
I can have a look at that later :)
https://rust-lang.github.io/mdBook/format/config.html

Wollac:
We can add custom styles later. I've created an issue for it so that we don't forget it.

PR #17, opened 2020-07-28
Add message rfc

GalRogozinski:
[Rendered](https://github.com/GalRogozinski/protocol-rfcs/blob/message/text/0017-message/0017-message.md)

### DIFF
--- a/text/0017-message/0017-message.md
+++ b/text/0017-message/0017-message.md
@@ -1,0 +1,180 @@
++ Feature name: `Tangle Messages`
++ Start date: 2020-07-28
++ RFC PR: [iotaledger/protocol-rfcs#0017](https://github.com/iotaledger/protocol-rfcs/pull/0017)
+
+# Summary
+
+A message is the object nodes gossip around in the network. It always references two other messages that are known as `parents`. It is stored as a vertex on the tangle data structure that the nodes maintain.
+
+The messages will contain payloads. Some of them will be core payloads that will be processed by all nodes as part of the core protocol. Some of them will be community payloads that will enable to build new functionality on top of the tangle.
+
+Some payloads may have other nested payloads embedded inside.
+So upon parsing it is done layer by layer.
+
+
+
+# Motivation
+
+To better understand this layered design, consider the internet protocol for example. There is an Ethernet frame, that contains an IP payload. This in turn contains a TCP packet that encapsulates an HTTP payload. Each layer has a certain responsibility, once this responsibility is completed, we move on to the next layer.
+
+The same goes with how we parse messages. The outer layer of the message enables us to map the message to a vertex in the Tangle and perform some basic validation. The next layer may be a transaction that mutates the ledger state. The next layer may provide some extra functionality on the transactions to be used by applications.
+
+By making it possible to add and exchange payloads, we are creating an architecture that can be easily extended to accommodate future needs.
+
+
+# Detailed design
+
+### Data Types
+
+The following are data types that we will use when we specify fields in the message and payloads.
+
+| Name   | Description   |
+| ------ | ------------- |
+| varint | An unsigned numerical value using a dynamic amount of bytes encoded in Little Endian. Evaluate by inspecting the MSB of the first byte. If it is `0` stop. If it is `1` go to the next byte. Continue applying this rule until you reach a byte with MSB `0`. Transform each byte to a 7-bit bit sequence by stripping the MSBs from each byte. Reverse the order of the 7 bit groups (Little Endian). Concatenate all the bits and evaluate as an unsigned integer. [It is similar to how it is done in Protocol Buffers](https://developers.google.com/protocol-buffers/docs/encoding). We must add [validation rules](#Message-Validation) that ensure that the varint is not malleable, and it is limited in size. |
+| uint64  | An unsigned 64 bit integer encoded in Little Endian.
+| ByteArray    | A dynamic size byte array. It is a prefixed by a `varint` that indicates the number of bytes in the array. The following bytes are simply the content of the array.
+| ByteArray[N] | A static size array of size N. Since the size is expected, no need to prefix with a varint   |
+
+
+### Message ID
+The message ID will be the `Blake2b-256` hash of the byte contents of the message. It should be used by the nodes to index the messages and by external APIs.
+
+
+### Message Structure
+
+<table>
+    <tr>
+        <th>Name</th>
+        <th>Type</th>
+        <th>Description</th>
+    </tr>
+    <tr>
+        <td>Version</td>
+        <td>varint</td>
+        <td>The message version. The schema specified in this RFC is for version <strong>1</strong> only. </td>
+    </tr>
+    <tr>
+        <td>Parent #1 (<code>trunk</code>)</td>
+        <td>ByteArray[32]</td>
+        <td>The Message ID of the first <i>Message</i> we reference.</td>
+    </tr>
+    <tr>
+        <td>Parent #2 (<code>branch</code>)</td>
+        <td>ByteArray[32]</td>
+        <td>The Message ID of the second <i>Message</i> we reference.</td>
+    </tr>
+    <tr>
+        <td>Payload Length</td>
+        <td>varint</td>
+        <td> The length of the Payload. Since its type may be unknown to the node it must be declared in advanced.</td>
+    </tr>
+    <tr>
+        <td colspan="1">
+            Payload
+        </td>
+        <td colspan="2">
+            <details open="true">
+                <summary>Generic Payload</summary>
+                <blockquote>
+                An outline of a general payload
+                </blockquote>
+                <table>
+                    <tr>
+                        <th>Name</th>
+                        <th>Type</th>
+                        <th>Description</th>
+                    </tr>
+                    <tr>
+                        <td>Payload Type</td>
+                        <td>varint</td>
+                        <td>
+                            The type of the payload. It will instruct the node how to parse the fields that follow. Types in the range of 0-127 are "core types" that all nodes are expected to know.
+                        </td>
+                    </tr>
+                    <tr>
+                        <td>Data Fields</td>
+                        <td>ANY</td>
+                        <td>A sequence of fields, where the structure depends on <code>payload type</code>.</td>
+                    </tr>
+                    <tr>
+                        <td>Nested Payload (Optional)</td>
+                        <td> Payload</td>
+                        <td> Some payloads support nesting other payloads inside of them.</td>
+                    </tr>
+                </table>
+            </details>
+    <tr>
+        <td>Nonce</td>
+        <td>uint64</td>
+        <td>The nonce which lets this message fulfill the Proof-of-Work requirement.</td>
+    </tr>
+</table>
+
+### Message PoW Hash
+
+This  hash is computed by converting all the messages raw bytes into trytes with `b1t6`. Then hash the result with CURL-P81. The result will be used to verify the proof-of-work done on the message. This will be used for spam protection.
+
+Ideally the Message ID would have been used. Then we could have fully deprecated all trinary remnants. However we must pay attention to specialized hardware that can render the PoW protection useless. The advantage of CURL that we have a scarce presence of FPGAs that can easily break it.
+
+### Message Validation
+
+A message is considered valid, if the following syntactic rules are met:
+
+1. The message length must not exceed X [tbd] bytes.
+2. Varint fields must adhere to the following rules:
+    * *Malleability protection:* No `0x80` or `0x00` trailing bytes. Take for example the following 3 byte varint: `0x8F8000`. 
+    In order to parse it we look at its bits: 
+    `1000 1111 1000 0000 0000 0000`.
+    First we strip the msbs:
+    `000 1111 000 0000 000 0000`
+    Then we reverse the groups of 7 bits (Little Endian):
+    `000 0000 000 0000 000 1111`
+    Then we concatenate all the bits to finally get:
+    `1111`
+    So the varint `0x0F` equals to `0x8F8000`, but is encoded with different bytes.
+    * *Size Protection*: The varint must not be more than 10 bytes long. A 10 byte varint can encompass all the range of `uint64`. In case of an overflow above the `uint64` range the message is invalid.
+2. When we are done parsing the message there shouldn't be any trailing bytes left that were not parsed.
+4. If the `payload type` is in the core payload range (0-127) and the node is familiar with it, or if it is in the range of 128-16383. The upper limit of 16383 is the maximal value of a 2 byte varint, and it should suffice for all the optional payload types we need.
+5. If the payload is valid according to the **syntactic** rules the `payload type` defines.
+6. If the Message PoW Hash will contain at least the number of trailing 0 trits the node defines as required.
+
+### Payloads
+
+Each message will contain a payload. The specification of the payloads is out of scope of this RFC. Below is a table of the currently specified core payloads with a link to their specifications. The `Unsigned Payload` will be specified here as an example.
+
+| Payload Name                            |
+| --------------------------------------- |
+| [Unsigned Data](#Unsigned-Data-Payload) |
+| Signed Data                             |
+| Transaction                             |
+| Indexation                              |
+
+### Unsigned Data Payload
+
+To make the Payload concept clear we will define the `unsigned data payload`. As the name suggests it simply allows to add arbitrary data to a message. A message that has been attached to the tangle and approved by a milestone has useful properties: You can verify that the content of the data did not change, and you can ascertain the approximate time it was published by checking the approving milestone. For example, one can record a hash of a text document on the tangle with such a payload. When a third party requests a proof that the document was published at a certain time and was not changed, one can point them to them to the message containing the `unsigned data payload` with the hash.
+
+The structure of the payload is simple:
+
+
+
+| Name             | Type     | Description               |
+| --------         | -------- | -----------               |
+| Payload Type     | varint   | Must be set to **2**      |
+| Data             | ByteArray| The data we are attaching |
+
+There are no validation rules for the payload. Message validation rules suffice here.
+
+
+
+# Rationale and alternatives
+
+Instead of creating a layered approach, we could have simply created a flat`transaction message`, that is tailored for mutating the ledger state, and try to fit all the use cases there. For example with the unsigned data use-case, we could have filled some section of the transaction with the data. Then via a flag in the transaction we could have instructed to not pass this transaction to the service that attempts to mutate the ledger state.
+
+This approach seems less extensible. It might have made sense if we wanted to build a protocol that is just for ledger mutating transactions, but we want to be able to extend the protocol to do more than that.
+
+# Unresolved questions
+
+- What should be the length of the message?
+- Better protection with PoW? how does the length of the message and type of payloads should affect difficulty?
+- How should we pad the message to fit it into CURL for PoW?
+- To what non-trinary hash function should we switch to?


text/0017-message/0017-message.md
```diff
@@ -0,0 +1,181 @@
...
+### Payloads
+
+Each message will contain a payload. The specification of the payloads is out of scope of this RFC. Below is a table of the currently specified core payloads with a link to their specifications. The `Unsigned Payload` will be specified here as an example.
+
+| Payload Name                              |   Type Value |
+| ---------------------------------------   | -----------  | 
+|  Signed Transaction                       |     0        |
+|  Signed Data                              |     1        |
+|  [Unsigned Data](#Unsigned-Data-Payload)  |     2        |
+| Indexation                                |     3        |
```
GalRogozinski, line 150: Type value can still be changed
This is just temporary... when I left it blank people asked kept on asking me why it is blank

I expect that in a meeting we will sort out the values

text/0017-message/0017-message.md
```diff
@@ -0,0 +1,181 @@
...
+Instead of creating a layered approach, we could have simply created a flat`transaction message`, that is tailored for mutating the ledger state, and try to fit all the use cases there. For example with the unsigned data use-case, we could have filled some section of the transaction with the data. Then via a flag in the transaction we could have instructed to not pass this transaction to the service that attempts to mutate the ledger state.
+
+This approach seems less extensible. It might have made sense if we wanted to build a protocol that is just for ledger mutating transactions, but we want to be able to extend the protocol to do more than that.
+
+# Unresolved questions
+
+- What should be the maximum length of the message?
+- Better protection with PoW? how does the length of the message and type of payloads should affect difficulty?
+- How should we pad the message to fit it into CURL for PoW?
+- How should we pad the input for CURL? One proposal is to go with padding method 2 that is described here https://en.wikipedia.org/wiki/ISO/IEC_9797-1#Padding
```
GalRogozinski, line 180: I was trying to get some response from researchers about this one but maybe I didn't do it hard enough

@Wollac offred the padding rule and I think it is good enough because even if someone manages to speed up PoW it won't be the most troubling thing

text/0017-message/0017-message.md
thibault-martinez: Is the Milestone payload missing though ?

text/0017-message/0017-message.md
```diff
@@ -0,0 +1,181 @@
...
+    First we strip the msbs:
+    `000 1111 000 0000 000 0000`
+    Then we reverse the groups of 7 bits (Little Endian):
+    `000 0000 000 0000 000 1111`
+    Then we concatenate all the bits to finally get:
+    `1111`
+    So the varint `0x0F` equals to `0x8F8000`, but is encoded with different bytes.
+    * *Size Protection*: The varint must not be more than 10 bytes long. A 10 byte varint can encompass all the range of `uint64`. In case of an overflow above the `uint64` range the message is invalid.
+2. When we are done parsing the message there shouldn't be any trailing bytes left that were not parsed.
+4. If the `payload type` is in the core payload range (0-127) and the node is familiar with it, or if it is in the range of 128-16383. The upper limit of 16383 is the maximal value of a 2 byte varint, and it should suffice for all the optional payload types we need.
```
cvarley100, line 137: How do register a custom (community) payload type?

Is 16383 sufficient for all applications?

text/0017-message/0017-message.md
```diff
@@ -0,0 +1,181 @@
...
+    <tr>
+        <td>Nonce</td>
+        <td>uint64</td>
+        <td>The nonce which lets this message fulfill the Proof-of-Work requirement.</td>
+    </tr>
+</table>
+
+### Message PoW Hash
+
+This hash is computed by converting all the messages raw bytes into trytes with `b1t6`. Then hash the result with Curl-P-81. The result will be used to verify the proof-of-work done on the message. This will be used for spam protection.
```
cvarley100, line 115: What is the reasoning behind continuing to use Curl-P-81?

GalRogozinski: Changing to a different PoW may open us up to spam attacks from people who possess specialized hardware.
With Curl-P-81 we know exactly what we are dealing with at the very least

We may switch from it later on after we have a conclusion to what PoW algo we want to switch to

text/0017-message/0017-message.md
GalRogozinski: IIRC, in our meetings we threw ideas on how it might be.
I think one idea was that community member will also create RFCs but not sure this is the best way to approach this

I think that together with community volunteers we should define at least one community payload. Then we should properly document and implement it, so other community members can easily follow.
We should probably talk about it more in a meeting.

Regarding the 16,383 (2^14 - 1) limit, I just thought it was a high enough number...
Given that we support colored coins it is hard for me to imagine that the payload feature will become so popular that this limit won't suffice...

text/0017-message/0017-message.md
```diff
@@ -0,0 +1,182 @@
...
+                </table>
+            </details>
+    <tr>
+        <td>Nonce</td>
+        <td>uint64</td>
+        <td>The nonce which lets this message fulfill the Proof-of-Work requirement.</td>
+    </tr>
+</table>
+
+### Message PoW Hash
```
Wollac, line 113: We need to explain this in more detail:
- We must have at least one example for the bytes of a serialized message.
- We need to carefully describe:
  - how the padding works
  - whether that is weighted by the length or something
  - how the difficulty is computed from the Curl hash
- Provide examples of PoWed data of different difficulty

text/0017-message/0017-message.md
```diff
@@ -0,0 +1,182 @@
...
+    First we strip the msbs:
+    `000 1111 000 0000 000 0000`
+    Then we reverse the groups of 7 bits (Little Endian):
+    `000 0000 000 0000 000 1111`
+    Then we concatenate all the bits to finally get:
+    `1111`
+    So the varint `0x0F` equals to `0x8F8000`, but is encoded with different bytes.
+    * *Size Protection*: The varint must not be more than 10 bytes long. A 10 byte varint can encompass all the range of `uint64`. In case of an overflow above the `uint64` range the message is invalid.
+2. When we are done parsing the message there shouldn't be any trailing bytes left that were not parsed.
+4. If the `payload type` is in the core payload range (0-127) and the node is familiar with it, or if it is in the range of 128-16383. The upper limit of 16383 is the maximal value of a 2 byte varint, and it should suffice for all the optional payload types we need.
```
Wollac, line 137: Limiting the number to 16383 or 2 bytes seems rather arbitrary. We have a varint anyways so way not use its power to potentially store much higher number when needed.

text/0017-message/0017-message.md
Wollac: We should really explain how we envision this community-driven payload draft process.
I guess we want to have a table somewhere where people can create a PR with a new ID and link to the document / RFC describing the payload.

Does it maybe make more sense to have a separate RFC with this table (similar to [SLIP-44](https://github.com/satoshilabs/slips/blob/master/slip-0044.md) or [SLIP-173](https://github.com/satoshilabs/slips/blob/master/slip-0173.md))? That could make the process considerably easier.

alexsporn: Why use a varint if you have a max bound of 2 bytes? It is way easier to define a uint16 here

text/0017-message/0017-message.md
```diff
@@ -0,0 +1,182 @@
...
+        <td>The Message ID of the first <i>Message</i> we reference.</td>
+    </tr>
+    <tr>
+        <td>Parent2 (<code>branch</code>)</td>
+        <td>ByteArray[32]</td>
+        <td>The Message ID of the second <i>Message</i> we reference.</td>
+    </tr>
+    <tr>
+        <td>Payload Length</td>
+        <td>varint</td>
```
alexsporn, line 68: What to we gain by using a varint here? Are you planning to have megabyte-sizes messages? Make it a uint32 and reduce complexity instead of making even the header of the message variable in size

text/0017-message/0017-message.md
```diff
@@ -0,0 +1,182 @@
...
+    `000 1111 000 0000 000 0000`
+    Then we reverse the groups of 7 bits (Little Endian):
+    `000 0000 000 0000 000 1111`
+    Then we concatenate all the bits to finally get:
+    `1111`
+    So the varint `0x0F` equals to `0x8F8000`, but is encoded with different bytes.
+    * *Size Protection*: The varint must not be more than 10 bytes long. A 10 byte varint can encompass all the range of `uint64`. In case of an overflow above the `uint64` range the message is invalid.
+2. When we are done parsing the message there shouldn't be any trailing bytes left that were not parsed.
+4. If the `payload type` is in the core payload range (0-127) and the node is familiar with it, or if it is in the range of 128-16383. The upper limit of 16383 is the maximal value of a 2 byte varint, and it should suffice for all the optional payload types we need.
+5. If the payload is valid according to the **syntactic** rules the `payload type` defines.
```
alexsporn, line 138: This actually breaks the layer concept. A message should be valid it **its** structure is valid, regardless of the payload.

text/0017-message/0017-message.md
GalRogozinski: We wanted to not have too many types in the message.
The idea is that less types means more simple descriptions and code.

Also, for lots of cases where we describe quantities `uint8` or `uint16` could have sufficed.

So the idea is that choosing varints will result mostly in shorter messages

text/0017-message/0017-message.md
GalRogozinski: See my other answer

Also Wolfgang seems to not like this bound, so maybe this bound will be removed (still need to discuss this with him)

text/0017-message/0017-message.md
GalRogozinski: Yup

This can actually be a problem the way it is defined here. If nodes will validate payloads differently (i.e. a newer node can know of a new payload type), this will affect the message...

I will remove this line

text/0017-message/0017-message.md
wusyong: I'm not convince it will be simpler code. It's not native primitives in many languages.
So we either implement one ourselves or use the existing one. Both would end up with more validation rules.

text/0017-message/0017-message.md
Wollac: I think the bound (for the non-core payloads) seems arbitrary and should be removed. We should allow for the full uint32 maybe even uint64 space.

text/0017-message/0017-message.md
oopsmonk: If `uint8_t` or `uint16_t` is sufficed, why not just use one of them? they are primitive types in most languages. The `varint` is good for data serialization but it is not easy to operate and memory management in a data structure, especially for microcontrollers.

GalRogozinski: due to recent conversations outside of github I am also leaning towards dropping varints

this decision will be finalized on the next protocol team meeting

text/0017-message/0017-message.md
```diff
@@ -0,0 +1,181 @@
++ Feature name: `Tangle Messages`
```
thibault-martinez, line 1:
```suggestion
+ Feature name: `tangle-message`
```
For consistency

text/0017-message/0017-message.md
```diff
@@ -0,0 +1,181 @@
...
++ RFC PR: [iotaledger/protocol-rfcs#0017](https://github.com/iotaledger/protocol-rfcs/pull/0017)
+
+# Summary
+
+A message is the object nodes gossip around in the network. It always references two other messages that are known as `parents`. It is stored as a vertex on the tangle data structure that the nodes maintain.
+
+The messages will contain payloads. Some of them will be core payloads that will be processed by all nodes as part of the core protocol. Some of them will be community payloads that will enable to build new functionality on top of the tangle.
+
+Some payloads may have other nested payloads embedded inside.
+So upon parsing it is done layer by layer.
```
thibault-martinez, line 12: Merge this sentence with the previous one.

text/0017-message/0017-message.md
```diff
@@ -0,0 +1,181 @@
...
+    </tr>
+    <tr>
+        <td>Version</td>
+        <td>varint</td>
+        <td>The message version. The schema specified in this RFC is for version <strong>1</strong> only. </td>
+    </tr>
+    <tr>
+        <td>Parent1 (<code>trunk</code>)</td>
+        <td>ByteArray[32]</td>
+        <td>The Message ID of the first <i>Message</i> we reference.</td>
```
thibault-martinez, line 59:
```suggestion
        <td>The Message ID of the first <i>Message</i> it references.</td>
```

text/0017-message/0017-message.md
```diff
@@ -0,0 +1,181 @@
...
+    </tr>
+    <tr>
+        <td>Parent2 (<code>branch</code>)</td>
+        <td>ByteArray[32]</td>
+        <td>The Message ID of the second <i>Message</i> we reference.</td>
```
thibault-martinez, line 64:
```suggestion
        <td>The Message ID of the second <i>Message</i> it references.</td>
```

text/0017-message/0017-message.md
thibault-martinez, line 115: We usually put the `t` before the `b`, so `t6b1`.

thibault-martinez: Also, trits, not trytes.

text/0017-message/0017-message.md
```diff
@@ -0,0 +1,181 @@
...
+    `1111`
+    So the varint `0x0F` equals to `0x8F8000`, but is encoded with different bytes.
+    * *Size Protection*: The varint must not be more than 10 bytes long. A 10 byte varint can encompass all the range of `uint64`. In case of an overflow above the `uint64` range the message is invalid.
+3. When we are done parsing the message there shouldn't be any trailing bytes left that were not parsed.
+4. If the `payload type` is in the core payload range (0-127) and the node is familiar with it, or if it is in the range of 128-16383. The upper limit of 16383 is the maximal value of a 2 byte varint, and it should suffice for all the optional payload types we need.
+5. If the Message PoW Hash will contain at least the number of trailing 0 trits the node defines as required.
+
+### Payloads
+
+Each message will contain a payload. The specification of the payloads is out of scope of this RFC. Below is a table of the currently specified core payloads with a link to their specifications. The `Unsigned Payload` will be specified here as an example.
```
thibault-martinez, line 142:
```suggestion
Each message will contain a payload. The specification of the payloads is out of scope of this RFC. Below is a table of the currently specified core payloads with a link to their specifications. The `Unsigned Data Payload` will be specified here as an example.
```

text/0017-message/0017-message.md
```diff
@@ -0,0 +1,181 @@
...
+| Indexation                                |     4        |
+
+### Unsigned Data Payload
+
+To make the Payload concept clear we will define the `unsigned data payload`. As the name suggests it simply allows to add arbitrary data to a message. A message that has been attached to the tangle and approved by a milestone has useful properties: You can verify that the content of the data did not change, and you can ascertain the approximate time it was published by checking the approving milestone. For example, one can record a hash of a text document on the tangle with such a payload. When a third party requests a proof that the document was published at a certain time and was not changed, one can point them to them to the message containing the `unsigned data payload` with the hash.
+
+The structure of the payload is simple:
+
+
+
```
thibault-martinez, line 159: Why 4 new lines :) ?

text/0017-message/0017-message.md
```diff
@@ -0,0 +1,181 @@
...
+| Name             | Type     | Description               |
+| --------         | -------- | -----------               |
+| Payload Type     | varint   | Must be set to **2**      |
+| Data             | ByteArray| The data we are attaching |
+
+There are no validation rules for the payload. Message validation rules suffice here.
+
+
+
```
thibault-martinez, line 168: Why 4 new lines :) ?

text/0017-message/0017-message.md
nothingismagick: I agree. ☝️ I see this artificial limitation as potentially shooting ourselves in the foot like with what happened during y2k.

text/0017-message/0017-message.md
Wollac: I think this is fundamentally different:
- `t5b1` encodes ternary data into binary, by converting each group of 5 trits into one byte; it cannot decode arbitrary binary data
- `b1t6` encodes binary data into ternary, by converting each byte into 6 trits; it cannot decode arbitrary ternary data

In [RFC-15](https://iotaledger.github.io/protocol-rfcs/0015-binary-to-ternary-encoding/0015-binary-to-ternary-encoding.html) it is also called `b1t6`.

Wollac: This is also motivated in #24

text/0017-message/0017-message.md
```diff
@@ -0,0 +1,181 @@
...
+                </table>
+            </details>
+    <tr>
+        <td>Nonce</td>
+        <td>uint64</td>
+        <td>The nonce which lets this message fulfill the Proof-of-Work requirement.</td>
+    </tr>
+</table>
+
+### Message PoW Hash
```
Wollac, line 113: The message PoW is now in its own RFC #24.
I think it would be best, to just remove this section. I wouldn't even reference #24 since this introduces a circular dependency and the exact PoW computation is not relevant in the context of this RFC.

text/0017-message/0017-message.md
```diff
@@ -0,0 +1,181 @@
...
+    `1111`
+    So the varint `0x0F` equals to `0x8F8000`, but is encoded with different bytes.
+    * *Size Protection*: The varint must not be more than 10 bytes long. A 10 byte varint can encompass all the range of `uint64`. In case of an overflow above the `uint64` range the message is invalid.
+3. When we are done parsing the message there shouldn't be any trailing bytes left that were not parsed.
+4. If the `payload type` is in the core payload range (0-127) and the node is familiar with it, or if it is in the range of 128-16383. The upper limit of 16383 is the maximal value of a 2 byte varint, and it should suffice for all the optional payload types we need.
+5. If the Message PoW Hash will contain at least the number of trailing 0 trits the node defines as required.
+
+### Payloads
+
+Each message will contain a payload. The specification of the payloads is out of scope of this RFC. Below is a table of the currently specified core payloads with a link to their specifications. The `Unsigned Payload` will be specified here as an example.
```
Wollac, line 142:
```
Each message will contain a payload.
```
Messages without payload data should be permitted. They will just have a payload length of zero but are otherwise 100% compatible to the specification.

Wollac (changes requested):

text/0017-message/0017-message.md
Wollac: Has been resolved with #24

text/0017-message/0017-message.md
GalRogozinski: I think maybe I can delete this entire section. Now we have an RFC dedicated for computing this PoW hash

text/0017-message/0017-message.md
```diff
@@ -0,0 +1,181 @@
...
+
+This approach seems less extensible. It might have made sense if we wanted to build a protocol that is just for ledger mutating transactions, but we want to be able to extend the protocol to do more than that.
+
+# Unresolved questions
+
+- What should be the maximum length of the message?
+- Better protection with PoW? how does the length of the message and type of payloads should affect difficulty?
+- How should we pad the message to fit it into CURL for PoW?
+- How should we pad the input for CURL? One proposal is to go with padding method 2 that is described here https://en.wikipedia.org/wiki/ISO/IEC_9797-1#Padding
+- Do we want to move away from using CURL for PoW in the future?
```
Wollac, line 181:
```suggestion
```
Should be removed as PoW is now discussed in #24

text/0017-message/0017-message.md
Wollac: Yes, the entire section should be deleted. See https://github.com/iotaledger/protocol-rfcs/pull/17#discussion_r476267452

GalRogozinski: due to me looking at the RFC while you are commenting, caused me to repeat your remarks

text/0017-message/0017-message.md
```diff
@@ -0,0 +1,181 @@
...
+The structure of the payload is simple:
+
+
+
+| Name             | Type     | Description               |
+| --------         | -------- | -----------               |
+| Payload Type     | varint   | Must be set to **2**      |
+| Data             | ByteArray| The data we are attaching |
+
+There are no validation rules for the payload. Message validation rules suffice here.
```
luca-moser, line 165: Does this mean that an `Unsigned Data Payload` with a byte array of length zero is valid?

text/0017-message/0017-message.md
GalRogozinski: I think it renders the same, but I will delete

text/0017-message/0017-message.md
GalRogozinski: I agree the message length protection suffices.
Actually, while thinking about the case for the bound, I started to question whether the message length is really needed...

text/0017-message/0017-message.md
GalRogozinski: If the `payload length` field in the message accommodates only for the type then in my opinion it should be valid.

Obviously there is no point in creating such a payload, but I don't see an issue here

text/0017-message/0017-message.md
karimodm: TL;DR: I would rather set fixed reasonable-sized types with some slack (you never know) replacing all varints.

`varints` are somewhat complex types to parse, but I would like to draw your attention to the creation of such message rather than the parsing of it. `varints` type force the entire message structure to "shift" instead of having fixed offsets. This is quite cumbersome when forming a message where all the parts are not necessarily known before-hand, and therefore the byte-requirements of length `varint` types is also unknown: when you fill in the data section of the message, you may see yourself push it forward in order to accommodate the corresponding `varint` size without overflowing into the data section itself. Sure, modern programming languages and libs provide abstract structures for all these, but at a cost of memory reallocation, memory concatenation and other forms of memory juggling. Leveling the participation field ultimately participates in the flexibility of a solution.

Thoralf-M:
What will prevent sending pruned messages again to spam the nodes?

GalRogozinski:
@Thoralf-M 
You have a point that before we could have dismissed old messages by looking at their timestamp and now we can't...
Sure this is a permissionless protocol, but we still have that PoW protection...

@Wollac maybe we should rethink how we approach timestamps?

GalRogozinski:
@Thoralf-M 
Since honest nodes won't gossip non-solid txs, someone who is circumventing PoW by broadcasting old txs might only be able to overload the nodes he is directly broadcasting to.

They won't spread to the rest of the network though...

Thoralf-M:
@GalRogozinski right that would help, don't know how big the attack vector is then, but broadcasting old messages seems like a really easy attack and if there are many nodes with auto peering and API open it must be considered in any case how the nodes handle this

text/0017-message/0017-message.md
```diff
@@ -0,0 +1,167 @@
...
+
+### Payloads
+
+A message may contain a payload. The specification of the payloads is out of scope of this RFC. Below is a table of the currently specified core payloads with a link to their specifications. The `indexation payload` will be specified here as an example.
+
+| Payload Name                              |   Type Value |
+| ---------------------------------------   | -----------  | 
+|  [Signed Transaction](https://github.com/luca-moser/protocol-rfcs/blob/signed-tx-payload/text/0000-signed-transaction-payload/0000-signed-transaction-payload.md)                       |     0        |
+|  [Milestone Draft](https://github.com/jakubcech/protocol-rfcs/blob/jakubcech-milestonepayload/text/0019-milestone-payload/0019-milestone-payload.md)                                |     1        |
+|  [Indexation Payload](#Unsigned-Data-Payload)  |     2        |
```
oopsmonk, line 123: I think it refer to indexation payload.
```suggestion
|  [Indexation Payload](#indexation-payload)  |     2        |
```

text/0017-message/0017-message.md
```diff
@@ -0,0 +1,167 @@
...
+        <td>ByteArray[32]</td>
+        <td>The Message ID of the first <i>Message</i> it references.</td>
+    </tr>
+    <tr>
+        <td>Parent2 (<code>branch</code>)</td>
+        <td>ByteArray[32]</td>
+        <td>The Message ID of the second <i>Message</i> it references.</td>
+    </tr>
+    <tr>
+        <td>Payload Length</td>
```
oopsmonk, line 65: is it memory length or serialization length of the payload?

GalRogozinski: serialization length

After deserializing the payload, in theory a node implementation can choose any way it wants to represent it in memory.
We attempt to spec the protocol, not the implementation details (unless they are necessary)

text/0017-message/0017-message.md
```diff
@@ -0,0 +1,172 @@
...
+There are no validation rules for the payload. Message validation rules suffice here.
+
+### Serialization Example
+
+Below is a serialized valid message with the indexation payload. The index is the "SPAM" ASCII string and the message is the "Hello Iota"
+ASCII string. The [Message PoW Hash](https://github.com/Wollac/protocol-rfcs/blob/message-pow/text/0024-message-pow/0024-message-pow.md) should have
+  14 trailing zeroes. Bytes are expressed as hexadecimal numbers. 
+
+[Version] **`01`** [Parent 1] `F532A53545103276B46876C473846D98648EE418468BCE76DF4868648DD73E5D` [Parent 2] `78D546B46AEC4557872139A48F66BC567687E8413578A14323548732358914A2` [Payload Length]
+*`0C`*[Payload Type]**`02`**[Index] *`04`*`9A52` [Data]*`0A`*`48656c6c6f20496f7461`[Nonce]`dc293f3333333333`
```
thibault-martinez, line 162: `9A52` == `SPAM` ?

GalRogozinski: Hmm...
Not sure how this happened :-|

Also the payload length is incorrect...
I will change the amount of trailing 0s to 10 for the example because it is easier to calculate

thibault-martinez: Also, the index length and data length seem to be only one byte ?

GalRogozinski: Before I fix, I am starting to wonder whether the 32 bytes we specified for the length is a good decision...
I understand it is convenient, but maybe it is just bloating up message size for no good reason?

text/0017-message/0017-message.md
```diff
@@ -0,0 +1,166 @@
...
+ASCII string. The [Message PoW Hash](https://github.com/Wollac/protocol-rfcs/blob/message-pow/text/0024-message-pow/0024-message-pow.md) would have
+  10 trailing zeroes for the given nonce in this example. Bytes are expressed as hexadecimal numbers.
+
+[Version] **`01`** [Parent 1] `F532A53545103276B46876C473846D98648EE418468BCE76DF4868648DD73E5D` [Parent 2] `78D546B46AEC4557872139A48F66BC567687E8413578A14323548732358914A2` [Payload Length]
+*`1100000000000000000000000000000000000000000000000000000000000000`*[Payload Type]**`02`**[Index] *`04`*`5350414D` [Data]*`0A`*`48656C6C6F20496F7461`[Nonce]`5d38333333333333`
+
+# Rationale and alternatives
+
+Instead of creating a layered approach, we could have simply created a flat `transaction message`, that is tailored for mutating the ledger state, and try to fit all the use cases there. For example with the unsigned data use-case, we could have filled some section of the transaction with the data. Then via a flag in the transaction we could have instructed to not pass this transaction to the service that attempts to mutate the ledger state.
```
charlesthompson3, line 160:
```suggestion
Instead of creating a layered approach, we could have simply created a flat `transaction message` that is tailored for mutating the ledger state, and try to fit all the use cases there. For example, with the unsigned data use-case, we could have filled some section of the transaction with the data. Then via a flag in the transaction, we could have instructed to not pass this transaction to the service that attempts to mutate the ledger state.
```

charlesthompson3 (commented):
Here are some minor edits to help the audience navigate the text a little faster; please let me know if you have any questions / concerns about the suggestions!

-Charles

text/0017-message/0017-message.md
```diff
@@ -0,0 +1,166 @@
++ Feature name: `tangle-message`
++ Start date: 2020-07-28
++ RFC PR: [iotaledger/protocol-rfcs#0017](https://github.com/iotaledger/protocol-rfcs/pull/0017)
+
+# Summary
+
+A message is the object nodes gossip around in the network. It always references two other messages that are known as `parents`. It is stored as a vertex on the tangle data structure that the nodes maintain.
+
+The messages will contain payloads. Some of them will be core payloads that will be processed by all nodes as part of the core protocol. Some of them will be community payloads that will enable to build new functionality on top of the tangle. Some payloads may have other nested payloads embedded inside.
```
charlesthompson3, line 9:
```suggestion
The messages will contain payloads. Some of them will be core payloads that will be processed by all nodes as part of the core protocol. Some of them will be community payloads that will enable to build new functionality on top of the Tangle. Some payloads may have other nested payloads embedded inside.
```

Thoralf-M:
Can we update it to 1-8 parents?

kuhlmannmarkus:
Im not sure about if thats a fact, but I think that the index for an indexation payload always has to be a valid UTF8 string (JS library). I dont understand why this is useful. I would like to be able to use plain bytes for the index. We are developing a solution that needs to have arbitraty indeces. The only way for us in that case would be bytes->Base64->UTF8. Again, I dont understand, why there is only UTF8 accepted on "protocol level".

GalRogozinski:
@luca-moser and @thibault-martinez 

IIRC, you guys pushed for enforcing UTF-8 in the implementation. It seemed reasonable to me, but now I honestly don't have a good explanation to give. Will something break if we stop enforcing UTF-8?

GalRogozinski:
@kuhlmannmarkus 
Just to understand your issue better, what are the sizes of the indexes you will use?

kuhlmannmarkus:
@GalRogozinski 
Thanks for picking this up. I would like to use a sha512 hash for the indices. So 64bytes.

GalRogozinski:
@kuhlmannmarkus 

In an internal discussion people said:
1. Keeping it as string is good for UX (i.e. dashboards) and most people will probably use strings.
2. People like yourself can use the base64 trick you mentioned.

They were also reluctant to make unnecessary changes.

Given those arguments I personally find it hard to pick a side and don't have much of a strong opinion of whether this change should be pushed.

If you have more arguments you can keep on writing them here. IF is reading :)

kuhlmannmarkus:
@GalRogozinski thanks!
I have some more implications on why to change the behaviour:

1. Using 64 bytes hashes will actually be impossible, since the corresponding base64 conversion will be 4/3 of the input size.
The UTF8 enforcement thus introduces a rather drastic restriction on the addressable space. I believe this is true for every general conversion from bytes to UTF8, not only for base64.

2. I dont really see the UX endangered. The actual contents of a message also have to be converted to bytes, so this is nothing unusual at this point. Users could still just convert their strings to bytes.

3. Applications like the explorer could try to parse the index anyways. The workload of checking the UTF8 integrity, while rather small, should not reside on the node side.

4. Human readable strings for the M2M economy? :-)

Stay healthy and all the best!

BenRoyce-Society2-com:
plain bytes does not seem like a difficult implementation

and the versatility seems worth it

there's a number of scenarios, besides markus's need, where it could become useful

kuhlmannmarkus:
awesome. thanks everyone!

text/0017-message/0017-message.md
```diff
@@ -0,0 +1,169 @@
...
+  A message that has been attached to the Tangle and approved by a milestone has several useful properties: verifying that the content of the data did not change and determining the approximate time it was published by checking the approving milestone. If the payload will be incorporated under
+  the `signed transaction payload`, the content will be signed as well.
+
+
+The structure of the payload is simple:
+
+| Name             | Type          | Description               |
+| --------         | -----------   | -----------               |
+| Payload Type     | uint32        | Must be set to **2**      |
+| Index            | ByteArray     | The index key of the message |
```
thibault-martinez, line 146: `ByteArray` has a length denoted by a `u32` but index should be denoted by a `u16`.

Thoralf-M:
Parent mesage ids need to be ordered and unique

text/0017-message/0017-message.md
```diff
@@ -0,0 +1,169 @@
...
+
+### Payloads
+
+A message may contain a payload. The specification of the payloads is out of scope of this RFC. Below is a table of the currently specified core payloads with a link to their specifications. The `indexation payload` will be specified here as an example:
+
+| Payload Name                              |   Type Value |
+| ---------------------------------------   | -----------  | 
+|  [Signed Transaction](https://github.com/luca-moser/protocol-rfcs/blob/signed-tx-payload/text/0000-signed-transaction-payload/0000-signed-transaction-payload.md)                       |     0        |
+|  [Milestone Draft](https://github.com/jakubcech/protocol-rfcs/blob/jakubcech-milestonepayload/text/0019-milestone-payload/0019-milestone-payload.md)                                |     1        |
+|  [Indexation Payload](#indexation-payload)  |     2        |
```
thibault-martinez, line 125: Should we add the 2 new payloads ?

text/0017-message/0017-message.md
```diff
@@ -0,0 +1,169 @@
...
+    </tr>
+    <tr>
+        <td>NetworkID</td>
+        <td>uint64</td>
+        <td>Network identifier. This field will signify whether this message was meant for mainnet, testnet, or a private net. It also tells what protocol rules apply to the message. It is first 8 bytes of the `BLAKE2b-256` hash of the concatenation of the network type and the protocol version string.</td>
+        </tr>
+    <tr>
+        <td> Parents' length </td>
+        <td> uint8</td>
+        <td> The number of messages we directly approve. Can be any value between 1-8.</td>
```
m-renaud, line 57: Could you clarify the exact range here a little bit? It's not immediately obvious if "between 1-8" is inclusive or exclusive (or half open), and in the first sentence of the summary it says "It always references two other messages that are known as `parents`".

Maybe something like: "Can be any value in the range (1,8]"?

text/0017-message/0017-message.md
```diff
@@ -0,0 +1,169 @@
...
+    </tr>
+    <tr>
+        <td>Parents </td>
+        <td>ByteArray[32 * `parents length`]</td>
+        <td>The Message IDs that are referenced.</td>
```
m-renaud, line 62: Could the description be expanded to say that these Message IDs are the ones directly approved? Currently that information is only in the "Parents' length" spec above.

text/0017-message/0017-message.md
```diff
@@ -0,0 +1,169 @@
...
+
+### Message validation
+
+A message is considered valid, if the following syntactic rules are met:
+
+1. The message size must not exceed 32 KiB (32 * 1024 bytes).
+2. When parsing the message is complete, there should not be any trailing bytes left that were not parsed.
+3. If the `payload type` is known to the node.
+4. If the [Message PoW Hash](https://github.com/Wollac/protocol-rfcs/blob/message-pow/text/0024-message-pow/0024-message-pow.md) will contain at least the number of trailing 0 trits the node defines as required.
+5. `Parents' length` must be between 1-8.
```
m-renaud, line 114: Same comment here re. clarification of allowed range.

text/0017-message/0017-message.md
```diff
@@ -0,0 +1,169 @@
...
+
+# Rationale and alternatives
+
+Instead of creating a layered approach, we could have simply created a flat `transaction message` that is tailored for mutating the ledger state, and try to fit all the use cases there. For example, with the unsigned data use-case, we could have filled some section of the transaction with the data. Then via a flag in the transaction, we could have instructed to not pass this transaction to the service that attempts to mutate the ledger state.
+
+This approach seems less extensible. It might have made sense if we wanted to build a protocol that is just for ledger mutating transactions, but we want to be able to extend the protocol to do more than that.
+
+# Unresolved questions
+
+- What should be the maximum length of the message?
```
m-renaud, line 169: Is this resolved? In the text above it says "The message size must not exceed 32 KiB (32 * 1024 bytes)"

text/0017-message/0017-message.md
```diff
@@ -0,0 +1,169 @@
++ Feature name: `tangle-message`
++ Start date: 2020-07-28
++ RFC PR: [iotaledger/protocol-rfcs#0017](https://github.com/iotaledger/protocol-rfcs/pull/0017)
+
+# Summary
+
+A message is the object nodes gossip around in the network. It always references two other messages that are known as `parents`. It is stored as a vertex on the tangle data structure that the nodes maintain.
+
+The messages will contain payloads. Some of them will be core payloads that will be processed by all nodes as part of the core protocol. Some of them will be community payloads that will enable the building of new functionalities on top of the Tangle. Some payloads may have other nested payloads embedded inside.
```
m-renaud, line 9: In the Payloads section below is says "A message _may_ contain a payload" but here it says "will", could you clarify if it is strictly required or not?

text/0017-message/0017-message.md
```diff
@@ -0,0 +1,169 @@
...
+1. The message size must not exceed 32 KiB (32 * 1024 bytes).
+2. When parsing the message is complete, there should not be any trailing bytes left that were not parsed.
+3. If the `payload type` is known to the node.
+4. If the [Message PoW Hash](https://github.com/Wollac/protocol-rfcs/blob/message-pow/text/0024-message-pow/0024-message-pow.md) will contain at least the number of trailing 0 trits the node defines as required.
+5. `Parents' length` must be between 1-8.
+
+
+### Payloads
+
+A message may contain a payload. The specification of the payloads is out of scope of this RFC. Below is a table of the currently specified core payloads with a link to their specifications. The `indexation payload` will be specified here as an example:
```
m-renaud, line 119: Here it say it "may" contain a payload, what would the message format be if no payload exists? It seems like by default the rest of the message would be zero'd out which could be interpreted as a Signed Transaction (according to the table below).

Wollac (approved):

text/0017-message/0017-message.md
```diff
@@ -0,0 +1,163 @@
++ Feature name: `tangle-message`
++ Start date: 2020-07-28
++ RFC PR: [iotaledger/protocol-rfcs#0017](https://github.com/iotaledger/protocol-rfcs/pull/0017)
+
+# Summary
+
+A message is the object nodes gossip around in the network. It always references two other messages that are known as _parents_. It is stored as a vertex on the tangle data structure that the nodes maintain.
```
thibault-martinez, line 7: > It always references two other messages that are known as _parents_.

Not factually correct

text/0017-message/0017-message.md
Wollac: done

Wollac (approved):

PR #18, opened 2020-07-28
RFC: Transaction Payload

luca-moser:
[Rendered Version](https://github.com/luca-moser/protocol-rfcs/blob/signed-tx-payload/text/0018-transaction-payload/0018-transaction-payload.md)


text/0000-signed-transaction-payload/0000-signed-transaction-payload.md
```diff
@@ -0,0 +1,825 @@
...
+    * `UTXO Input`:
+        * `Transaction Output Index` must be 0 ≤ x < 127.
+        * Every combination of `Transaction ID` + `Transaction Output Index` must be unique in the inputs set.
+    * Inputs must be in lexicographical order of their serialized form.<sup>1</sup>
+* Outputs:
+    * `Outputs Count` must be 0 < x < 127.
+    * At least one output must be specified.
+    * `Output Type` must be 0, denoting a `SigLockedSingleDeposit`.
+    * `SigLockedSingleDeposit`:
+        * `Address Type` must either be 0 or 1, denoting an `Ed25519`- or `WOTS address`.
```
thibault-martinez, line 755: Reverse for consistency

text/0000-signed-transaction-payload/0000-signed-transaction-payload.md
```diff
@@ -0,0 +1,825 @@
...
+
+| Index | Must Contain                                                                                                         |
+| ----- | ---------------------------------------------------------------------------------------------------------------- |
+| 0     | A <i>Signature Unlock Block</i> holding the corresponding Ed25519 signature to unlock A and C.                    |
+| 1     | A <i>Signature Unlock Block</i> which unlocks B.                                                                 |
+| 2     | A <i>Reference Unlock Block</i> which references index 0, since C also gets unlocked by the same signature as A. |
+
+## Validation
+
+A <i>Signed Transaction</i> payload has different validation stages, since some validation steps can only be executed if certain information becomes present, respectively should have become present. We therefore distinguish between syntactical- and semantic validation.
```
cvarley100, line 734: The phrasing here is unclear.

luca-moser: Can you elaborate what is unclear?

text/0000-signed-transaction-payload/0000-signed-transaction-payload.md
```diff
@@ -0,0 +1,825 @@
...
+                    <tr>
+                        <td>Transaction Type</td>
+                        <td>varint</td>
+                        <td>
+                        Set to <strong>value 0</strong> to denote an <i>Unsigned Transaction</i>.
+                        </td>
+                    </tr>
+                   <tr>
+                        <td>Inputs Count</td>
+                        <td>varint</td>
```
alexsporn, line 118: why use varint if you set the rule to be < 127? why not uint8?

text/0000-signed-transaction-payload/0000-signed-transaction-payload.md
```diff
@@ -0,0 +1,825 @@
...
+                                        <td>varint</td>
+                                        <td>The index of the output on the referenced transaction to consume.</td>
+                                    </tr>
+                                </table>
+                            </details>
+                        </td>
+                    </tr>
+                   <tr>
+                        <td>Outputs Count</td>
+                        <td>varint</td>
```
alexsporn, line 158: why use varint if you set the rule to be < 127? why not uint8?

text/0000-signed-transaction-payload/0000-signed-transaction-payload.md
```diff
@@ -0,0 +1,825 @@
...
+                                        </td>
+                                    </tr>
+                                    <tr>
+                                        <td>Transaction ID</td>
+                                        <td>ByteArray[32]</td>
+                                        <td>The BLAKE2b-256 hash of the transaction from which the UTXO comes from.</td>
+                                    </tr>
+                                    <tr>
+                                        <td>Transaction Output Index</td>
+                                        <td>varint</td>
```
alexsporn, line 149: why use varint if you set the rule to be < 127? why not uint8?

text/0000-signed-transaction-payload/0000-signed-transaction-payload.md
```diff
@@ -0,0 +1,825 @@
...
+                                </blockquote>
+                                <table>
+                                    <tr>
+                                        <th>Name</th>
+                                        <th>Type</th>
+                                        <th>Description</th>
+                                    </tr>
+                                    <tr>
+                                        <td>Payload Type</td>
+                                        <td>varint</td>
```
alexsporn, line 323: why use varint if you set the rule to be < 127? why not uint8?

text/0000-signed-transaction-payload/0000-signed-transaction-payload.md
```diff
@@ -0,0 +1,825 @@
...
+                    <tr>
+                        <td>Unlock Type</td>
+                        <td>varint</td>
+                        <td>
+                            Set to <strong>value 1</strong> to denote a <i>Reference Unlock Block</i>.
+                        </td>
+                    </tr>
+                    <tr>
+                        <td>Reference</td>
+                        <td>varint</td>
```
alexsporn, line 443: why use varint if you set the rule to be < Count (max 127)? why not uint8?

text/0000-signed-transaction-payload/0000-signed-transaction-payload.md
Wollac: This should clearly be consistent with the _Payload_ as introduced in #17. Since this representsthe layered payload approach.

There is no rule limiting this to 127 much larger values should be allowed.

text/0000-signed-transaction-payload/0000-signed-transaction-payload.md
Wollac: I am afraid that 127 might be too limiting for the allowed number of inputs/outputs.

It might be worth considering a fixed integer type for this field to assure that an input block always has a fixed size.

text/0000-signed-transaction-payload/0000-signed-transaction-payload.md
Wollac: Using a fixed integer type here could give us a fixed byte size for the reference block, if that is desirable.

text/0000-signed-transaction-payload/0000-signed-transaction-payload.md
Wollac: I think, the type of a length/count is a slightly different situation than a regular field. They represent serialized arrays of any kind and here always consistently serializing them by a prepended varint makes sense.
On of the main arguments of fixed integer types is not valid in this context as the length of the entire struct will always be dynamic due to the array.

text/0000-signed-transaction-payload/0000-signed-transaction-payload.md
```diff
@@ -0,0 +1,825 @@
...
+                                        <td>Signature Type</td>
+                                        <td>varint</td>
+                                        <td>
+                                            Set to <strong>value 0</strong> to denote a <i>WOTS Signature</i>.
+                                        </td>
+                                    </tr>
+                                    <tr>
+                                        <td>Signature</td>
+                                        <td>ByteArray</td>
+                                        <td>The signature signing the serialized <i>Unsigned Transaction</i>.</td>
```
wusyong, line 388: How to sign this with wots signature scheme using Kerl? AFAIK input message of kerl still have to be represented in trits.

text/0000-signed-transaction-payload/0000-signed-transaction-payload.md
```diff
@@ -0,0 +1,825 @@
...
+                            </details>
+                        </td>
+                    </tr>
+                    <tr>
+                        <td>Payload Length</td>
+                        <td>varint</td>
+                        <td>The length in bytes of the optional payload.</td>
+                    </tr>
+                   <tr>
+                        <td valign="top">Payload <code>optOneOf</code></td>
```
laumair, line 246: Can we support multiple payloads? For example, if one wants to pass a tag and also some unsigned data?

Wollac: The general idea would be to have a new payload type `PayloadChain` or something that supports something like this. However, in order to keep the very initial payload types to a minimum it has not been spec'ed yet.

text/0000-signed-transaction-payload/0000-signed-transaction-payload.md
Wollac: The legacy W-OTS just expects a 243-trit hash to be signed. One could use plain Keccak-384 for this and then convert the 48-byte hash to trits (exactly as it is currently done as part of Kerl). Then the current version of the ISS can be used exactly.

There will be a new RFC in the next days that describes this in detail, but as hinted above it is really not that complicated.

text/0000-signed-transaction-payload/0000-signed-transaction-payload.md
laumair: @Wollac Sounds good.

text/0000-signed-transaction-payload/0000-signed-transaction-payload.md
wusyong: Could I just convert the message to trits and use the current implementation of WOTS? If we use Keccak-384 that means another signing scheme implementation. It's not complicated for sure, but that doesn't mean it's not redundant.

Wollac: No you cannot, you need exactly 243 trits. If you convert the Message using `b1t6` you get an arbitrary amount of trits.
The method I've described above is exactly Kerl, just the first step that converts the input is skipped. That's also exactly why I've mentioned Keccak-384 (which is used for Kerl) and not some other hash function, to avoid the introduction of some new hashing function.

text/0000-signed-transaction-payload/0000-signed-transaction-payload.md
```diff
@@ -0,0 +1,825 @@
...
+                                                    <tr>
+                                                        <td><b>Name</b></td>
+                                                        <td><b>Type</b></td>
+                                                        <td><b>Description</b></td>
+                                                    </tr>
+                                                    <tr>
+                                                        <td>Address Type</td>
+                                                        <td>varint</td>
+                                                        <td>
+                                                            Set to <strong>value 0</strong> to denote an <i>Ed25519 Address</i>.
```
Thoralf-M, line 219: Value should be 1

text/0000-signed-transaction-payload/0000-signed-transaction-payload.md
frohal: Maybe I'm wrong here, but - When I sign a WOTS-signature - the only thing I actually use, is 1-3 (sec-level) balanced arrays of ints from -13 to 13. - aka "The'normalized bundle-hash" .. When creating a new signature-algo-rithm - it should be enough to 'recreate' this setup using any kind of algorithm - and o avoid th M-bug - one should avoid having the first valuee of any of the 3 set to 13. There should not be a need for exactly 243 trits at all- The trits used during signing, is from the private key, What is neede to sign is therefore only a balanced array so that the sum of each of the three sub-arrays is 0.

Wollac: > What I'm saying, is that the 'input-message' to be signed is an int-array of 81 ints ranging from -13 to 13.
Exactly, that's what I meant above! You need 81 trytes (-13...13) as an input. 81 trytes are exactly the 243 trits I've mentioned.

It is debatable whether the normalization is part of the signature scheme or part of the generation of the "input-message" to be signed. I'd usually consider this as part of the signature scheme as this is very special to the IOTA W-OTS and not used in other Winternitz signature schemes, but this is just a definition.

frohal: The normalisation-step used today, is only a security-step to avoid exposing more than 50% of the key.
Signing avd validation fo the signature based on a pure bundle-hash would work just fine, but would be less secure, since you could easily end up exposing more than 80% of the key if the bundle-has was 'bad' for signing-purpose.
In that regard - however you chosse to calcualte those three parts to be signed - you should keep the normalisation as a security-spep, but you don't have to do it the same way when creating a new signature-schema. For instance - you could choose to have the new 'input' as a 24-based value (0-23) and normalize on 12 (or 11), and hash 4 more times to verify the public key (address) and that would only occure on the nodes when validating the signature.  24 is a more suidted number for binary operations - you would however make sure that the values ectracted to sign needs to be converted to 81 numbers from 0 to 23. This would be bit less secure that using all the 27 hashings for signing, but probably less expensive to calculate on the node.

frohal: Actually - if you choose to use 24-based signing-schema - you could choose to start with 2 hashes before starting on the input - thus immideatly avoiding the M-bug, since you will never push the first signature-part as a clean copy of the private key :)

Wollac: > The normalisation-step used today, is only a security-step to avoid exposing more than 50% of the key.

Unfortunately, that is not correct. The normalization is a crucial and integral part to the security of the used W-OTS signature scheme. Without it, after the (first) signature has been exposed, an attacker has the signature for every bundle hash that is element-wise ≥ than the signed bundle hash. Even with just one signature, it would be trivial for an attacker to produce valid signatures for other messages. With normalization this is not possible. Traditional W-OTS schemes use a checksum that decreases when any element of the input increases. This has exactly the same effect, but it makes the signatures even longer and is therefore not 100% compatible with the current scheme.

However, you are right: How exactly the normalization is performed can be changed while still retaining 100% compatibility.
For example: Currently, the normalization even favors `M`s at the beginning of the bundle hash and this could be switched.
Also, you are of course correct that you can always make the codomain of the function used for the bundle hash smaller and still get a valid input for the signing. The main drawback with this is, that you (potentially considerably) reduce the security of the signature scheme: By reducing the codomain, it will be easier to find collisions, maybe even, because of some strange interplay of hash and normalization, considerably easier than just the reduction factor....

Using base24 is an interesting idea :thinking: although I am not really sure how much it would actually speed up the verification process as the bundle hash computation is only a small part of it.
One even faster alternative would for example be to consider each set bit of the hash as a 1/-1 and each un-set bit as a 0. However, this would reduce the security of the (not normalized) bundle hash from 243-trits to 243 bits...

Wollac: We will soon be publishing an RFC on Github that explains the "binary W-OTS" which we are discussing in more detail and provide alternatives.
@frohal, your feedback and ideas on those would then of course be greatly appreciated.

wusyong: I hope the "Binary-WOTS" could entirely based on Keccak. I was trying to implement the scheme with first step skipped (which is we doesn't convert trits to binary in Kerl). And it turned out to be a lot of works. Both WOTS and Kerl in bee would have to redesign since they are only considered ternary originally.

The new scheme I hope could be is we just need to convert the ternary seed to binary and just use Keccak do the hashing. So only Wots layer would need to rewrite.

karimodm: > One could use plain Keccak-384 for this and then convert the 48-byte hash to trits (exactly as it is currently done as part of Kerl)
Just a clarification here: you won't be able to use `b1t6` on the resulting Keccak-384 squeeze as it would yield 288 trits, unusable for signing. `iota.go`'s -> `kerl` -> `KerlBytesToTrits` will be used instead; do we have it documented somewhere?

text/0000-signed-transaction-payload/0000-signed-transaction-payload.md
```diff
@@ -0,0 +1,825 @@
...
+                            <td>The signature signing the serialized <i>Unsigned Transaction</i>.</td>
+                        </tr>
+                    </table>
+                </details>
+            </td>
+        </tr>
+    </table>
+</details>
+
+A <i>Signature Unlock Block</i> defines an <i>Unlock Block</i> which holds one or more signatures unlocking one or more inputs. Such block signs the entire <i>Unsigned Transaction</i> part of a <i>Signed Transaction</i> including the optional payload.
```
karimodm, line 694: The wording here  is misleading: a single Unlock Block contains only one signature; this signature in turns **may** be used to unlock more than its corresponding one input only if referenced by a proceeding Reference Unlock Block.

text/0000-signed-transaction-payload/0000-signed-transaction-payload.md
```diff
@@ -0,0 +1,825 @@
...
+                                        <td>uint64</td>
+                                        <td>The amount of tokens to deposit with this <i>SigLockedSingleDeposit</i> output.</td>
+                                    </tr>
+                                </table>
+                            </details>
+                        </td>
+                    </tr>
+                    <tr>
+                        <td>Payload Length</td>
+                        <td>varint</td>
```
karimodm, line 242: I am not a fan of these `varint` nested structures: when there is an outer `varint` that defines the length in bytes of a structure containing a `varint`. As the inner varint changes in size so might the outer, possibly creating a cascading effect as we expand the protocol further.
Of course this logic does not apply to `varint` count fields.

text/0000-signed-transaction-payload/0000-signed-transaction-payload.md
```diff
@@ -0,0 +1,825 @@
...
+
+##### Payload
+
+The payload part of an <i>Unsigned Transaction</i> can hold an optional payload. This payload does not affect the validity of the <i>Unsigned Transaction</i>. If the transaction is not valid, then the payload must also be discarded.
+
+Supported payload types to be embedded into an <i>Unsigned Transaction</i>:
+
+| Name                  | Type Value |
+| --------------------- | ---------- |
+| Unsigned Data Payload | 2          |
```
Thoralf-M, line 593: There should only be the Indexation Payload now?

frohal:
Hi all - I assume that when this is first deployed, the complete balance on a address will be put into one UTXO ?
And when spending from a WOTS-address (like all have to do the first time they move funds) they would preferably send the complete output - Would the transaction ve evaluated to 'invalid' i fnot all funds in one UTXO is actually accounted for in the outputs ? - Like - you have 5K in one address (now UTXO) and want to send 1K, you would now have to send 4k back to own address making a new UTXO for 4K ?  Today one can just 'leave' the funds at the address, but that would probably no longer be possible ?

This would then immideatly avoid all the currenlty possible replay-thefts - right ?

text/0000-signed-transaction-payload/0000-signed-transaction-payload.md
```diff
@@ -0,0 +1,742 @@
...
+                                        </td>
+                                    </tr>
+                                    <tr>
+                                        <td>Transaction ID</td>
+                                        <td>ByteArray[32]</td>
+                                        <td>The BLAKE2b-256 hash of the transaction from which the UTXO comes from.</td>
+                                    </tr>
+                                    <tr>
+                                        <td>Transaction Output Index</td>
+                                        <td>uint16</td>
```
wusyong, line 149: The index is in range of 0 ≤ x < 127. So I guess uint8 is more than enough?

luca-moser: Of course it would be enough to use a uint8 but we aren't yet sure whether we're actually going to limit by 127 or not.

text/0000-transaction-payload/0000-transaction-payload.md
```diff
@@ -0,0 +1,748 @@
...
+
+Using an UTXO based model provides several benefits:
+* Parallel validation of transactions.
+* Easier double-spend detection, since conflicting transactions would reference the same UTXO.
+* Replay-protection which is especially important when having reusable addresses. Replaying the same transaction would manifest itself as already being applied or existent and thus not have any impact.
+* Technically seen, balances may no longer be associated to addresses which raises the level of abstraction and thus enables other types of outputs. Consider for example a type of output which specifies the balance to be unlocked by a transaction which must fulfill a very hard Proof-of-Work difficulty or supply some other unlock criteria etc.
+
+Within a transaction using UTXOs, inputs and outputs make up the to be signed data of the transaction. The section unlocking the inputs is called *unlock block*. An unlock block may contain a signature proving ownership of a given input's address and/or other unlock criteria.
+
+The following image depicts the flow of funds using UTXO:
```
frohal, line 40: Will there be one 'genesis' transaction for each address with balance at the time of convertion to new tx-layout ?

text/0000-signed-transaction-payload/0000-signed-transaction-payload.md
frohal: @Wollac I might be folling this on the wrong places, but I am not sure I can find that RFC you said that would be soon published.  Care to share a link to it ?

charlesthompson3 (commented):
Here are some edits to help the audience navigate the text a little faster; please let me know if you have any questions / concerns about the suggestions!

-Charles

text/0000-transaction-payload/0000-transaction-payload.md
```diff
@@ -0,0 +1,748 @@
...
+
+A <i>Transaction Essence</i> must contain at least one input and output.
+
+##### Inputs
+
+The <i>Inputs</i> part holds the inputs to consume, respectively, to fund the outputs of the <i>Transaction Essence</i>. There is only one type of input as of now, the <i>UTXO Input</i>. In the future, more types of inputs may be specified as part of protocol upgrades.
+
+Each defined input must be accompanied by a corresponding <i>Unlock Block</i> at the same index in the <i>Unlock Blocks</i> part of the <i>Transaction Payload</i>. 
+If multiple inputs can be unlocked through the same <i>Unlock Block</i>, then the given <i>Unlock Block</i> only needs to be specified at the index of the first input which gets unlocked by it. 
+Subsequent inputs which are unlocked through the same data must have a <i>Reference Unlock Block</i> pointing to the previous <i>Unlock Block</i>.
```
karimodm, line 384:
```suggestion
Subsequent inputs which are unlocked through the same data must have a <i>Reference Unlock Block</i> pointing to the index of a previous <i>Unlock Block</i>. 
```

karimodm: Ambigous interpretation fixed.

karimodm (changes requested):

Thoralf-M:
Should all WOTS stuff be removed?

frohal:
> Should all WOTS stuff be removed?

Doesen't the next version of the ledger and thereby the transactions need to support WOTS-addresses at all ? How would the funds get moved from WOTS-addresses to new address-format if one should not sign TX'es from WOTS to move them ?

Or are you thinking that there should be a 'special' transaction-paylload only for moving funds from a WOTS-address, and that the new ledger would not support transactions _to_ WOTS-addresses at all, and that this RFC should only consider tx'es based on new address-format ?  In that case - there would also be need for a spec on the 'transition-tx' needed to move from WOTS

luca-moser:
There will be further RFC changes to reflect the route we're taking re WOTS.

text/0000-transaction-payload/0000-transaction-payload.md
```diff
@@ -0,0 +1,748 @@
...
+This validation can commence as soon as the transaction data has been received in its entirety. It validates the structure but not the signatures of the transaction. If the transaction does not pass this stage, it must not be broadcasted further and can be discarded right away.
+
+The following criteria defines whether the transaction passes the syntactical validation:
+* `Transaction Essence Type` value must be 0, denoting an `Transaction Essence`.
+* Inputs:
+    * `Inputs Count` must be 0 < x < 127.
+    * At least one input must be specified.
+    * `Input Type` value must be 0, denoting an `UTXO Input`.
+    * `UTXO Input`:
+        * `Transaction Output Index` must be 0 ≤ x < 127.
```
thibault-martinez, line 668: Input/output count is `0 < x < 127` -> 126 possible.
Output index is `0 ≤ x < 127` -> 127 possible.

Wollac (approved):

text/0000-transaction-payload/0000-transaction-payload.md
luca-moser: This process will be described in a subsequent RFC.

text/0000-transaction-payload/0000-transaction-payload.md
```diff
@@ -0,0 +1,749 @@
...
+                                                        <td>The raw bytes of the Ed25519 address which is a BLAKE2b-256 hash of the Ed25519 public key.</td>
+                                                    </tr>
+                                                </table>
+                                            </details>
+                                        </td>
+                                    </tr>
+                                    <tr>
+                                        <td>Amount</td>
+                                        <td>uint64</td>
+                                        <td>The amount of tokens to deposit with this <i>SigLockedSingleOutput</i> output.</td>
```
Thoralf-M, line 235: I don't see that it can't be 0, shouldn't that info be added here?

luca-moser: Check out the "Syntactical Validation" section.

oopsmonk:
#32 SigLockedDustAllowanceOutput is missing in `outputs` and `Syntactical Validation` section.

text/0000-transaction-payload/0000-transaction-payload.md
```diff
@@ -0,0 +1,562 @@
...
+A <i>Transaction Payload</i> payload is made up of two parts:
+1. The <i>The Transaction Essence</i> part which contains the inputs, outputs and an optional embedded payload.
+2. The <i>Unlock Blocks</i> which unlock the <i>Transaction Essence</i>'s inputs. In case the unlock block contains a signature, it signs the entire <i>Transaction Essence</i> part.
+
+All values are serialized in little-endian encoding. The serialized form of the transaction is deterministic, meaning the same logical transaction always results in the same serialized byte sequence.
+
+A `Blake2b-256` hash of the entire serialized data makes up <i>Transaction Payload</i>'s ID.
+
+Following table structure describes the entirety of a <i>Transaction Payload</i>'s serialized form:
+* [Data Type Notation](https://github.com/GalRogozinski/protocol-rfcs/blob/message/text/0017-message/0017-message.md#data-types)
```
m-renaud, line 59: This seems to be referenced in several places, would it make sense to have an RFC to define the datatypes which would be the canonical location instead of RFCs cross linking each other?

Wollac: I agree. However since this effects several RFCs it should happen in a separate RFC. I'll create an issue for that.

Wollac (approved):

lzpap (approved):

karimodm (approved):

PR #19, opened 2020-07-28
RFC: Milestone Payload

jakubcech:


cvarley100:
Could a non-Quantum robust signature scheme lead to any issues with the historical ledger state with the advent of QCs?

text/0019-milestone-payload/0019-milestone-payload.md
```diff
@@ -0,0 +1,58 @@
...
+- The Ed25519 signature function takes a Ed25519 private key, a sequence of bytes of arbitrary size and produces an Ed25519 signature of length 64 bytes. The given sequence of bytes should then be internally hashed (using sha512) by the same function.
+- The Ed25519 verification function takes a public key, a sequence of bytes of arbitrary size, a Ed25519 signature, and returns true/false based on the signature validity.
+
+To generate a valid milestone, the Coordinator *MUST*: 
+1. Generate a *Message* as defined in [RFC-0017 (draft)](https://github.com/GalRogozinski/protocol-rfcs/blob/message/text/0017-message/0017-message.md).
+2. Generate a new [milestone payload](#Milestone-payload) without filling the signature field.
+3. Sign the serialized bytes given by the concatenation of the following fields:
+    - Version, Parent1, Parent2, Payload Length of the Message;
```
alexsporn, line 27: I understand why you do this, but it actually breaks the "layer" concept of the messages.

Wollac: I agree, this is certainly not perfect. However, I'd also say that it is not that dramatic as this references only a layer down (and not up which would indeed be a complete paradigm shift).

For a milestone, we need to make 100% sure that the branch/trunk (and probably also the message type) cannot be forged. 

@Alex6323 do you have any alternative idea to achieve this?
Copying those fields into the MS payload also does not seem very good.

text/0019-milestone-payload/0019-milestone-payload.md
```diff
@@ -0,0 +1,58 @@
...
+    - Payload Type, Index Number, Timestamp, Inclusion Merkle Proof of the Milestone Payload.
```
thibault-martinez, line 28:
```suggestion
    - `Payload Type`, `Index Number`, `Timestamp`, `Inclusion Merkle Proof` of the Milestone Payload.
```

text/0019-milestone-payload/0019-milestone-payload.md
```diff
@@ -0,0 +1,58 @@
...
+
+# Unresolved questions
+
+- Should we add support for multi-signature or multi-stage signature? Could that be a desired feature from a devOps perspective?
+- Are we sure we want to lose quantum-robustness? We could have used a hash-based signature scheme, such as [XMSS](https://tools.ietf.org/html/rfc8391) or [LMS](https://tools.ietf.org/html/rfc8554) that provide quantum robustness at the price of increasing both communication and computation overhead. For more detail, please refer to this [document](https://docs.google.com/document/d/15_FkOhHFR4arxBBl07H_ETUGjPbf5jlJOiyYwZ7zKOg/edit?usp=sharing).
+- Do we want to pick Ed25519 or Ed448? They provide a security level of 128-bit and 224-bit respectively. Size of both private and public keys are of 32 or 57 bytes depending on the curve used. Similarly, the signature size is of 64 or 114 bytes. Ed25519 has better library-wise support with respect to Ed448.
+- Should we add a Network ID field to the payload? If yes, is the ID a string or a uint64?
```
thibault-martinez, line 58: What is this field again ?

text/0019-milestone-payload/0019-milestone-payload.md
```diff
@@ -0,0 +1,58 @@
...
+| Index Number           | uint64          | The index number of the milestone.                                                                                                                                                                                                                                                                      |
+| Timestamp              | uint64          | The Unix timestamp at which the milestone was issued. The unix timestamp is specified in seconds.                                                                                                                                                                                                       |
+| Inclusion Merkle Proof | Array<byte>[64] | Specifies the merkle proof which is computed out of all the tail transaction hashes of all the newly confirmed state-mutating bundles. ([RFC-0012](https://github.com/iotaledger/protocol-rfcs/blob/master/text/0012-milestone-merkle-validation/0012-milestone-merkle-validation.md)) |
```
thibault-martinez, line 46: No more tails and bundles :)

text/0019-milestone-payload/0019-milestone-payload.md
Wollac: There is no a cleaner separation of layers, as the Milestone Payload contains the reference parent1 and parent2.

text/0019-milestone-payload/0019-milestone-payload.md
Wollac: fixed

text/0019-milestone-payload/0019-milestone-payload.md
```diff
@@ -0,0 +1,136 @@
...
+  - The signature provider service will sign the received serialized bytes as-is.
+  - The signature provider will serialize the signature bytes and return them to the Coordinator.
+- Fill the `Signatures` field of the milestone payload with the received signature bytes.
+- Generate a *Message* as defined in [RFC-0017 (draft)](https://github.com/GalRogozinski/protocol-rfcs/blob/message/text/0017-message/0017-message.md) using the same `Parent1` and `Parent2` for the created _Milestone Payload_.
+
+## Syntactical validation
+
+- `Parent1` and `Parent2` of the payload must match `Parent1` and `Parent2` of the encapsulating _Message_.
+- `Keys Count` must be at least the _Signature Threshold_ and at most the number of _Applicable Public Keys_ for the current milestone index.
+- Public keys:
```
karimodm, line 123: We need to add something important here: the keys **must** not repeat.

karimodm (changes requested):

text/0019-milestone-payload/0019-milestone-payload.md
```diff
@@ -0,0 +1,136 @@
...
+
+In the IOTA protocol, nodes use the milestones issued by the Coordinator to reach a consensus on which transactions are confirmed. 
+This RFC proposes a new milestone payload as well as the EdDSA signature algorithm as a replacement to the current Coordinator digital signature scheme to authenticate the issued milestones.
+
+# Motivation
+The current signature scheme used to authenticate milestones issued by the Coordinator has been designed with ternary and quantum robustness in mind. Due to the switch to binary, the new message layout as well as the upcoming Coordicide (i.e., Coordinator removal), we now have the opportunity to revisit such a mechanism and replace it with a more simple, well-vetted and standardized one. Although the proposed digital signature scheme, EdDSA, does not provide quantum robustness, its simplicity and easier key-managment make it a good candidate for the time being.
+
+# Detailed design
+
+The EdDSA signature algorithm in its variants **Ed25519** and **Ed448** (providing a security level of 128-bit and 224-bit respectively) are technically described in the IRTF [RFC 8032](https://tools.ietf.org/html/rfc8032).
```
GalRogozinski, line 16: If we won't be using ed448 let's not mention it here
I just find it confusing

text/0019-milestone-payload/0019-milestone-payload.md
```diff
@@ -0,0 +1,136 @@
...
+Size of both private and public keys are of 32 or 57 bytes depending on the curve used. Similarly, the signature size is of 64 or 114 bytes.
+
+- The Ed25519 key generation can be fed with a given random sequence of bytes (i.e., seed) of length 32 bytes. Output of the key generation is the pair of private and public keys. Note that the generation of the seed is out of the scope of this RFC, but in general, any cryptographic pseudo-random function (PRF) would suffice.
+- The Ed25519 signature function takes a Ed25519 private key, a sequence of bytes of arbitrary size and produces an Ed25519 signature of length 64 bytes. The given sequence of bytes should then be internally hashed (using sha512) by the same function.
```
GalRogozinski, line 20: I understand that you meant that you simply hash the message before you sign it

But the way it is phrased is not clear to me... "same function" reads as though it refers to ed25519 signature function which is not sha-512

I guess by "internally hashed" you meant the ed25519 impl is an object that has an internal hash function?

it confused me for real

text/0019-milestone-payload/0019-milestone-payload.md
```diff
@@ -0,0 +1,136 @@
...
+
+# Rationale and alternatives
+
+Instead of going with EdDSA we could have chosen ECDSA. Both algorithms are well supported and widespread. However, signing with ECDSA requires fresh randomness while EdDSA does not. Moreover, we could have used a commit-reveal mechanism to update and commit the Coordinator public key at each next milestone. This method would make a quantum-based attack aimed at breaking the "current" Coordinator private-key more difficult. On the other hand, key management as well as verification of the milestone chain would become more complex.
+
+# Unresolved questions
+
+- Forcing matching `Parent1`, `Parent2` in the _Milestone Payload_ and its _Message_ makes it impossible to reattach the same payload at different positions in the Tangle. While this does not prevent reattachments in general (a different, valid `Nonce`, for example would lead to a new Message ID), this still simplifies milestone processing. However, it violates a clean separation of payload and message. As such, it might still be desirable to slightly complicate the milestone processing in order to allow any reattachments of _Milestone Payloads_ by not validating the parents.
```
GalRogozinski, line 136: "in order to allow"?

we must not allow reattachments

text/0019-milestone-payload/0019-milestone-payload.md
Wollac: Added an explicit line that the public keys must be unique

text/0019-milestone-payload/0019-milestone-payload.md
```diff
@@ -0,0 +1,137 @@
...
++ RFC PR: [iotaledger/protocol-rfcs#0019](https://github.com/iotaledger/protocol-rfcs/pull/19)
++ Author: Angelo Capossele
+
+# Summary
+
+In the IOTA protocol, nodes use the milestones issued by the Coordinator to reach a consensus on which transactions are confirmed. 
+This RFC proposes a new milestone payload as well as the EdDSA signature algorithm as a replacement to the current Coordinator digital signature scheme to authenticate the issued milestones.
+
+# Motivation
+The current signature scheme used to authenticate milestones issued by the Coordinator has been designed with ternary and quantum robustness in mind. Due to the switch to binary, the new message layout as well as the upcoming Coordicide (i.e., Coordinator removal), we now have the opportunity to revisit such a mechanism and replace it with a more simple, well-vetted and standardized one. Although the proposed digital signature scheme, EdDSA, does not provide quantum robustness, its simplicity and easier key-managment make it a good candidate for the time being.
```
charlesthompson3, line 12:
```suggestion
# Motivation
The current signature scheme used to authenticate milestones issued by the Coordinator has been designed with ternary and quantum robustness in mind. Due to the switch to binary, the new message layout, and the upcoming Coordicide (i.e., Coordinator removal), we now have the opportunity to revisit such a mechanism and replace it with a simpler, well-vetted and standardized one. Although the proposed digital signature scheme, EdDSA, does not provide quantum robustness, its simplicity and easier key-managment make it a good candidate for the time being.
```

text/0019-milestone-payload/0019-milestone-payload.md
```diff
@@ -0,0 +1,137 @@
...
+
+# Detailed design
+
+The EdDSA signature algorithm in its variants **Ed25519** and **Ed448** (providing a security level of 128-bit and 224-bit respectively) are technically described in the IRTF [RFC 8032](https://tools.ietf.org/html/rfc8032).
+Size of both private and public keys are of 32 or 57 bytes depending on the curve used. Similarly, the signature size is of 64 or 114 bytes.
+
+- The Ed25519 key generation can be fed with a given random sequence of bytes (i.e., seed) of length 32 bytes. Output of the key generation is the pair of private and public keys. Note that the generation of the seed is out of the scope of this RFC, but in general, any cryptographic pseudo-random function (PRF) would suffice.
+- The Ed25519 signature function takes a Ed25519 private key, a sequence of bytes of arbitrary size and produces an Ed25519 signature of length 64 bytes. The given sequence of bytes should then be internally hashed (using sha512) by the same function.
```
charlesthompson3, line 20:
```suggestion
The size of both private and public keys are roughly 32 or 57 bytes depending on the curve used. Similarly, the signature size is 64 or 114 bytes.

- The Ed25519 key generation can be fed with a random given sequence of bytes (i.e., seed) the length of 32 bytes. Output for the key generation is the pair of private and public keys. Note that the generation of the seed is out of the scope of this RFC, but in general, any cryptographic pseudo-random function (PRF) would suffice.
- The Ed25519 signature function takes a Ed25519 private key, a sequence of bytes of arbitrary size and produces an Ed25519 signature length of 64 bytes. The given sequence of bytes should then be internally hashed (using sha512) by the same function.
```

text/0019-milestone-payload/0019-milestone-payload.md
```diff
@@ -0,0 +1,137 @@
...
+- The Ed25519 verification function takes a public key, a sequence of bytes of arbitrary size, a Ed25519 signature, and returns true/false based on the signature validity.
+
+In order to increase the security of the design, a milestone can optionally be independently signed by multiple keys at once. These keys should be operated by detached signature provider services, running on independent infrastructure elements, thus mitigating the risk of an attacker having access to all the key material necessary for forging milestones. While the Coordinator takes responsibility of forming Milestone Payload Messages, it delegates signing to these providers through an ad-hoc RPC connector. Mutual authentication should be enforced between the Coordinator and the signature providers: a [client-authenticated TLS handshake](https://en.wikipedia.org/wiki/Transport_Layer_Security#Client-authenticated_TLS_handshake) scheme is advisable. To increase the flexibility of the mechanism, nodes can be configured to require a quorum of valid signatures to consider a milestone as genuine.
```
charlesthompson3, line 23:
```suggestion
To increase the security of the design, a milestone can (optionally) be independently signed by multiple keys at once. These keys should be operated by detached signature provider services running on independent infrastructure elements. This assist in mitigating the risk of an attacker having access to all the key material necessary for forging milestones. While the Coordinator takes responsibility for forming Milestone Payload Messages, it delegates signing in to these providers through an ad-hoc RPC connector. Mutual authentication should be enforced between the Coordinator and the signature providers: a [client-authenticated TLS handshake](https://en.wikipedia.org/wiki/Transport_Layer_Security#Client-authenticated_TLS_handshake) scheme is advisable. To increase the flexibility of the mechanism, nodes can be configured to require a quorum of valid signatures to consider a milestone as genuine.
```

charlesthompson3 (commented):
Here are some minor edits for clarity within the content; please let me know if you have any questions / concerns about the suggestions!

text/0019-milestone-payload/0019-milestone-payload.md
Wollac: Has been added and explained in #17

text/0019-milestone-payload/0019-milestone-payload.md
```diff
@@ -0,0 +1,58 @@
++ Feature name: `Milestone Payload`
++ Start date: 2020-07-28
++ RFC PR: [iotaledger/protocol-rfcs#0019](https://github.com/iotaledger/protocol-rfcs/pull/19)
++ Author: Angelo Capossele
+
+# Summary
+
+In the IOTA protocol, nodes use the milestones issued by the Coordinator to reach a consensus on which transactions are confirmed. 
+This RFC proposes a new milestone payload as well as the EdDSA signature algorithm as a replacement to the current Coordinator digital signature scheme to authenticate the issued milestones.
+
```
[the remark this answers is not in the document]
Wollac: I agree. I've changed the text accordingly.

text/0019-milestone-payload/0019-milestone-payload.md
Wollac: removed

text/0019-milestone-payload/0019-milestone-payload.md
Wollac: I've removed this paragraph completely as it does add much information.

text/0019-milestone-payload/0019-milestone-payload.md
Wollac: Thanks

text/0019-milestone-payload/0019-milestone-payload.md
Wollac: It seems to be practically infeasible/impossible to completely prevent reattachments. As such, it is much easier and cleaner to consider the Milestone as a "virtual marker" located at the corresponding position than an actual message location. I've adapted the RFC to clarify this aspect.

Wollac (approved):

Wollac:
@capossele since I am not sure how much of your original proposal is left, I've just requested a quick review from you :sweat_smile:

capossele (approved):

karimodm (approved):

text/0019-milestone-payload/0019-milestone-payload.md
```diff
@@ -0,0 +1,133 @@
...
++ Author: Angelo Capossele
+
+# Summary
+
+In IOTA, nodes use the milestones issued by the Coordinator to reach a consensus on which transactions are confirmed. This RFC proposes a milestone payload for the messages described in [Draft RFC-17](https://github.com/GalRogozinski/protocol-rfcs/blob/message/text/0017-message/0017-message.md). It uses Edwards-curve Digital Signature Algorithm (EdDSA) to authenticate the milestones.
+
+# Motivation
+
+In the current IOTA protocol, milestones are authenticated using a ternary Merkle signature scheme. With [Chrysalis](https://roadmap.iota.org/chrysalis), ternary transactions will be replaced with with binary messages containing different payload types. In order to address these new requirements, this RFC proposes the use of a dedicated payload type for milestones. It contains the same essential data fields that were previously included in the milestone bundle. Additionally, this document also describes how Ed25519 signatures are used to assure authenticity of the issued milestones. In order to make the management and security of the used private keys easier, simple multisignature features with support for key rotation have been added.
```
thibault-martinez, line 12:
```suggestion
In the current IOTA protocol, milestones are authenticated using a ternary Merkle signature scheme. With [Chrysalis](https://roadmap.iota.org/chrysalis), ternary transactions will be replaced with binary messages containing different payload types. In order to address these new requirements, this RFC proposes the use of a dedicated payload type for milestones. It contains the same essential data fields that were previously included in the milestone bundle. Additionally, this document also describes how Ed25519 signatures are used to assure authenticity of the issued milestones. In order to make the management and security of the used private keys easier, simple multisignature features with support for key rotation have been added.
```

text/0019-milestone-payload/0019-milestone-payload.md
```diff
@@ -0,0 +1,133 @@
...
+            <td>uint32</td>
+            <td>The index number of the milestone.</td>
+          </tr>
+          <tr>
+            <td>Timestamp</td>
+            <td>uint64</td>
+            <td>The Unix timestamp at which the milestone was issued. The unix timestamp is specified in seconds.</td>
+          </tr>
+          <tr>
+            <td>Parent1</td>
```
thibault-martinez, line 65: Need to add multiple parent support

thibault-martinez (changes requested):

PR #20, opened 2020-07-28
RFC: Bech32 Address Format

Wollac:


text/0020-bech32-address-format/0020-bech32-address-format.md
```diff
@@ -0,0 +1,87 @@
...
+Currently, two kinds of addresses are supported:
+ - W-OTS, which is compatible to the existing legacy IOTA signature scheme and
+ - Ed25519, where the address consists of the BLAKE2b-256 hash of the Ed25519 public key.
+
+They are serialized as follows:
+
+| Type    | First byte | Address bytes                                             |
+| ------- | ---------- | --------------------------------------------------------- |
+| W-OTS   | `0x00`     | 49 bytes: The t5b1 encoded Kerl hash of the W-OTS digest. |
+| Ed25519 | `0x01`     | 32 bytes: The BLAKE2b-256 hash of the Ed25519 public key. |
```
wusyong, line 40: Does this mean we are going to use all things defined in Ed25 including keypair generation instead of 243 tris or whatever we use before?

Wollac: I am not sure I fully understand this comment.

In the current proposal RFC https://github.com/iotaledger/protocol-rfcs/pull/18 there are two alternative signature schemes that can be used:
1. Legacy W-OTS 100% compatible to existing private keys, addresses and thus based on 243-trit hashes.
2. Ed25519 exactly as described in [RFC-8032](https://tools.ietf.org/html/rfc8032), which is based entirely on binary private- / public keys.

wusyong: thanks, the second one about if Ed25 is entirely based on binary key pair is what I want to know.

cvarley100:
An HRP of `iota` would likely be preferable from a UX point of view.

Wollac:
@cvarley100 the issue is that Bech32 addresses should not be longer than 90 chars. Legacy W-OTS addresses require 80 chars (49 bytes). This plus separator (1 char) plus checksum (6 chars) only leaves up to 3 chars for the HRP.

Instead of violating this restriction of the Bech32 specification (and thus considerably weakening the error detection/correction), there are also some tricks to compress the W-OTS addresses down to 48 bytes, which would allow up to 4 char HRP. But this would require additional description, specs and also probably implementation of this mechanism.

Unless there are some strong UX/PR reasons to go for `iota` it would be easier to settle with a 3-char HRP like `iot`.

cvarley100:
@Wollac Why do we need a separator?

Wollac:
That is just how the [Bech32](https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki) standard is defined.
They even give a motivation for this here: https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki#cite_note-3

jakubcech:
All for the 3 char `iot` HRP

mrbianchi:
Is good for UX having equal HRP for WOTS and ED25519? ( I know WOTS is larger )

Wollac:
> Is good for UX having equal HRP for WOTS and ED25519?

Traditionally, there is only one HRP per coin. This is exactly how other cryptos are handling it and I believe that this is also the intention behind the Bech32 spec:
- [BIP-173](https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki): P2WPKH and P2WSH use the same HRP (they only differ in length); different witness versions are only denoted by the first byte after the separator.
- [SLIP-173](https://github.com/satoshilabs/slips/blob/master/slip-0173.md): You can only register one HRP per coin and Mainnet / Testnet.

On compromise could be to encode the `address type` in 5 bits instead of 8 bits (as described in the current version). This would be a bit closer to how  [BIP-173](https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki) encodes the witness version and it would result into having the first character after the separator exactly denote the address type, i.e. `iot1q` would denote W-OTS and `iot1p` would denote Ed25519 etc...
However, encoding the version in 5 bits seems a bit arbitrary and would also introduce the complexity of the implementation a bit as we are then mixing base32 with base256. As the completely different address length already provide a very strong visual indicator, I am not sure this would be worth it.

> And will be good for UX changing the actual legacy address format for a new one?

The address format described above is just a different encoding for the addresses. No information is lost during conversation. In theory, the wallet could still stick with the legacy format or convert it internally.
That being said, the legacy tryte-based format is not very elegant as it does not offer any version / type, uses a more ambiguous charset and uses a much much weaker checksum not providing any guarantees and easier malleability. Especially since the Bech32 address format has support for different types it would have been strange to not also define it for W-OTS.

text/0020-bech32-address-format/0020-bech32-address-format.md
```diff
@@ -0,0 +1,87 @@
...
+## Examples
+
+- **Mainnet**
+   - W-OTS address hash (81-tryte): `EQSAUZXULTTYZCLNJNTXQTQHOMOFZERHTCGTXOLTVAHKSA9OGAZDEKECURBRIXIJWNPFCQIOVFVVXJVD9`
+      - serialized (50-byte): `00ea38caaeeec5c3fd6d0e89e32c9097bcc02ddcec40a8dcc516c7bc012d0b6bbe315493b0b86cdfbd11a7b8f0d5aa6f2200`
+      - Bech32 string: `iot1qr4r3j4wamzu8ltdp6y7xtysj77vqtwua3q23hx9zmrmcqfdpd4muv25jwctsmxlh5g60w8s6k4x7gsq28c8da`
+   - Ed25519 address hash (32-byte): `52fdfc072182654f163f5f0f9a621d729566c74d10037c4d7bbb0407d1e2c649`
+      - serialized (33-byte): `0152fdfc072182654f163f5f0f9a621d729566c74d10037c4d7bbb0407d1e2c649`
+      - Bech32 string: `iot1q9f0mlq8yxpx2nck8a0slxnzr4ef2ek8f5gqxlzd0wasgp73utryjtzcp98`
+- **Devnet**
```
thibault-martinez, line 65: Inconsistent use of `Testnet` and `Devnet`

Wollac: Yep! Someone proposed to change it from `Devnet` to general `Testnet`. Must have missed that one. Thanks

luca-moser (approved):

cvarley100:
@Wollac I really think we should go with `iota`. We have a four letter name, we should take advantage of that. It is only the legacy WOTS addresses that would be over 90 chars. 

I propose instead that we use a different identifier for wots, to differentiate it, and allow us to use `iota` for ed25.

Wollac:
I don't think this is an option. Bech32 and SLIP-0173 are standards that say that addresses must be at most 90 characters and that they have one human-readable denominator per coin. We should not break those requirements.

@cvarley100 you have mentioned two aspects in your comment:
- If we really want to make the different address types more distinguishable (in case the length difference of 61 vs 90 is not obvious enough) we should modify the "version byte" to make it more readable; e.g. Ed25519 as `iot1a` and WOTS as `iot1b`. However
- If we really want to the denominator to be 4-char `iota`, we have exactly two options:
  - No longer support WOTS in the protocol. 
  - Store the WOTS address as at most 48 bytes. This is possible, by either truncating the address hash - this will theoretically lead to more address collisions! Or by encode the address as 48 bytes by applying the [Kerl trits-bytes encoding](https://github.com/iotaledger/kerl/blob/master/IOTA-Kerl-spec.md#trits---bytes-encoding). Currently, this step is only part of the Kerl hash function and implementing this encoding would require additional afford ad well as validation etc. Does having a slightly nicer hrp really justify these steps?

cvarley100:
@Wollac Why do legacy addresses need to subscribe to bech32 encoding?

Wollac:
Of course they don't need to be, they can be in Morse code or any other encoding. However, it is exactly the point of this RFC to have one common encoding for all current and future address types and versions.
If it is really that crucial to exactly have `iota`, then we should either remove WOTS from the protocol completely or we need to bite the bullet and implement the more complicated encoding.

jakubcech:
I'm not a fan of going to great lengths for the iota HRP. Especially since neither of the complications to achieve 48 byte address has are very nice, unless the collision rate would be absolutely negligible. I see iota being better for our UX for sure, but I don't see it as that big a benefit. 

At the end of the day, we're building a protocol for the IoT. 🤷

text/0020-bech32-address-format/0020-bech32-address-format.md
```diff
@@ -0,0 +1,87 @@
...
+The human readable encoding of the address is Bech32 (as described in [BIP-0173](https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki)). A Bech32 string is at most 90 characters long and consists of: 
+
+- The **human-readable part** (HRP), which conveys the IOTA protocol and distinguishes between Mainnet (the IOTA token) and Testnet (testing version):
+   -  `iot` is the human-readable part for Mainnet addresses
+   -  `tio` is the human-readable part for Testnet addresses
+- The **separator**, which is always `1`.
+- The **data part**, which consists of the Base32 encoded serialized address and the 6-character checksum.
+
+For legacy W-OTS, the data part will consist of 86 characters (80 characters for the 50-byte address in Base32 and 6 characters for the checksum). Together with the HRP, separator and checksum, this leads to a Bech32 string of exactly 90 characters.
+Ed25519-based addresses will result in a Bech32 string of 63 characters.
```
charlesthompson3, line 54:
```suggestion
The human-readable encoding of the address is Bech32 (as described in [BIP-0173](https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki)). A Bech32 string is at most 90 characters long and consists of: 

- The **human-readable part** (HRP), which conveys the IOTA protocol and distinguishes between Mainnet (the IOTA token) and Testnet (testing version):
   -  `iot` is the human-readable part for Mainnet addresses
   -  `tio` is the human-readable part for Testnet addresses
- The **separator**, which is always `1`.
- The **data part**, which consists of the Base32 encoded serialized address and the 6-character checksum.

For legacy W-OTS, the data part will consist of 86 characters (80 characters for the 50-byte address in Base32 and 6 characters for the checksum). Together with the HRP, separator and checksum, this leads to a Bech32 string of exactly 90 characters.
Ed25519-based addresses will result in a Bech32 string of 63 characters.
```

charlesthompson3 (commented):
Here are a pair of very minor edits to grammar; please let me know if you have any questions about them!

-Charles

karimodm:
After conversations with wallet people, various engineers and bizdev the most agreed is 4-char prefix: `iota` / `atoi`.

text/0020-bech32-address-format/0020-bech32-address-format.md
```diff
@@ -0,0 +1,78 @@
...
+   - The remaining bytes contain the type-specific raw address bytes.
+
+Currently, only one kind of addresses are supported:
+ - Ed25519, where the address consists of the BLAKE2b-256 hash of the Ed25519 public key.
+
+They are serialized as follows:
+
+| Type    | First byte | Address bytes                                             |
+| ------- | ---------- | --------------------------------------------------------- |
+| Ed25519 | `0x01`     | 32 bytes: The BLAKE2b-256 hash of the Ed25519 public key. |
```
Wollac, line 38: The first and only supported address type (Ed25519) uses `1` instead of `0`. This tries to stay consistent with the address type in other RFCs, but should be unified once those types have been finalized.

Wollac:
@cvarley100 you can make the final call about the prefixes and we will adapt those!

thibault-martinez:
@Wollac Can you remind me why we kept type=1 for ED25 ?

text/0020-bech32-address-format/0020-bech32-address-format.md
thibault-martinez: Should we just go forward, remove WOTS from the other RFCs and use 0 ?

Wollac:
OK `iota` and `atoi` it is!

PR #21, opened 2020-07-29
Fix order of the purple set

JakeSCahill:
The RFC describes an ordering algorithm based on depth-first search, where you travel down the trunk of a milestone first, then down the branch.

I tried the example, and I think maybe O and N were the wrong way round.

So you would:

- Apply I
- Go back to O
- Apply K
- Apply O
- Go back to S
- Apply N
- Apply S
- Apply V

### DIFF
--- a/text/0005-white-flag/0005-white-flag.md
+++ b/text/0005-white-flag/0005-white-flag.md
@@ -131,7 +131,7 @@ The set of blue bundles `{Q, U, X, Y, Z, W, T, P}` is confirmed by another miles
 ![][Tangle]
 
 Applying the previously shown algorithm on the purple set produces the topological order
-`{D, G, J, L, M, R, I, K, N O, S, V}`.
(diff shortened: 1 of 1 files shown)


thibault-martinez:
Hmm, you don't go back to `0` after applying `I`, you go back to `K` ?
You always want to go trunk first then branch, so from `V` it would be (skipping the upper part with trunk) `V`->`S`->`N`->`K`->`I`.

JakeSCahill:
Where would O go?

thibault-martinez:
You access O through S, and O is the branch of S, so N (as trunk of S) has to come first

JakeSCahill:
Like in my changes?
~`D, G, J, L, M, R, I, K, N, O, S, V`~
`{D, G, J, L, M, R, I, K, O, N, S, V}`

thibault-martinez:
No, not like in your change. In your change `O` comes first. Why does O comes first when it's a branch and we say trunk have to come first ?

PR #25, opened 2020-08-26
RFC: Local snapshot file format

luca-moser:
[Rendered](https://github.com/luca-moser/protocol-rfcs/blob/local-snapshot-file-format/text/0000-local-snapshot-file-format/0000-local-snapshot-file-format.md)
[Sample code](https://github.com/luca-moser/iota/blob/master/local_snapshot_file_test.go)

### DIFF
--- a/text/0000-local-snapshot-file-format/0000-local-snapshot-file-format.md
+++ b/text/0000-local-snapshot-file-format/0000-local-snapshot-file-format.md
@@ -1,8 +1,3 @@
----
-title: Local Snapshot File Format
-tags: RFC, draft
----
-
 + Feature name: `local_snapshot_file_format`
 + Start date: 2020-08-25
 + RFC PR: [iotaledger/protocol-rfcs#0000](https://github.com/iotaledger/protocol-rfcs/pull/0000)
--- a/text/0000-local-snapshot-file-format/0000-local-snapshot-file-format.md
+++ b/text/0000-local-snapshot-file-format/0000-local-snapshot-file-format.md
@@ -172,5 +172,7 @@ Unlike the current format, this new format does no longer include:
 
 # Unresolved questions
 
-* Can we get rid of the spent addresses or do we still need to keep them?
 * Is all the information to startup a node from the local snapshot available with the described format?
+* Can we get rid of the spent addresses or do we still need to keep them?
+* Can we just use a byte for the index or do we need to use something different?
+* Do we need to account for different types of outputs already? (we currently only have them deposit to addresses)
--- a/text/0000-local-snapshot-file-format/0000-local-snapshot-file-format.md
+++ b/text/0000-local-snapshot-file-format/0000-local-snapshot-file-format.md
@@ -1,0 +1,181 @@
+---
+title: Local Snapshot File Format
+tags: RFC, draft
+---
+
++ Feature name: `local_snapshot_file_format`
++ Start date: 2020-08-25
++ RFC PR: [iotaledger/protocol-rfcs#0000](https://github.com/iotaledger/protocol-rfcs/pull/0000)
+
+# Summary
+
+This RFC defines a file format for local snapshots which is compatible with Chrysalis Phase 2.
+
+# Motivation
+
+Nodes create local snapshots to produce ledger representations at a point in time of a given milestone in order to be able to:
+* start up from a recent milestone instead of having to synchronize from genesis
+* delete transaction data below the given milestone
+
+Current node implementations use a [local snapshot file format](https://github.com/iotaledger/iri-ls-sa-merger/tree/351020d3b5e342b6e9a41f2868575ab7ff8c251c#generating-an-export-file-from-a-localsnapshots-db) which only works with account based ledgers. For Chrysalis Phase 2 this file format has to be assimilated to support a UTXO based ledger.
+
+# Detailed design
+
+All types are serialized in little-endian and occur in the sequence of the rows defined below. Local snapshot files are compressed via zlib to further reduce size.
+
+SEP = solid entry point. `Array<T>` are prefixed with a varint denoting the length.
+
+<table>
+    <tr>
+        <th>Name</th>
+        <th>Type</b></th>
+        <th>Description</th>
+    </tr>
+    <tr>
+        <td>Milestone Index</td>
+        <td>uint64</td>
+        <td>
+        The index of the milestone of the local snapshot.
+        </td>
+    </tr>
+    <tr>
+        <td>Milestone Hash</td>
+        <td>ByteArray[32]</td>
+        <td>
+        The BLAKE2b-256 hash of the milestone payload.
+        </td>
+    </tr>
+    <tr>
+        <td>Timestamp</td>
+        <td>uint64</td>
+        <td>
+        The UNIX epoch timestamp in seconds of when this snapshot was created.
+        </td>
+    </tr>
+    <tr>
+        <td>SEPs</td>
+        <td>Array<ByteArray[32]></td>
+        <td>
+        The BLAKE2b-256 hashes of the SEP messages at the cut off point of the given milestone.
+        </td>
+    </tr>
+    <tr>
+        <td>UTXOs</td>
+        <td colspan="2">
+            <details open="true">
+                <summary>Array<UTXO></summary>
+                <blockquote>
+                Describes the unspent transaction outputs per transaction.
+                </blockquote>
+                <table>
+                    <tr>
+                        <td><b>Name<b></td>
+                        <td><b>Type</b></td>
+                        <td><b>Description</b></td>
+                    </tr>
+                    <tr>
+                        <td>Transaction Hash</td>
+                        <td>ByteArray[32]</td>
+                        <td>The BLAKE2b-256 hash of the transaction.</td>
+                    </tr>
+                    <tr>
+                        <td>
+                        Unspent Outputs
+                        </td>
+                        <td colspan="2">
+                            <details open="true">
+                                <summary>Array<Outputs></summary>
+                                <table>
+                                    <tr>
+                                        <td>Index</td>
+                                        <td>byte</td>
+                                        <td>The index of the output on the transaction.</td>
+                                    </tr>
+                                    <tr>
+                                        <td valign="top">Address <code>oneOf</code></td>
+                                        <td colspan="2">
+                                            <details>
+                                                <summary>WOTS Address</summary>
+                                                <table>
+                                                    <tr>
+                                                        <td><b>Name<b></td>
+                                                        <td><b>Type</b></td>
+                                                        <td><b>Description</b></td>
+                                                    </tr>
+                                                    <tr>
+                                                        <td>Address Type</td>
+                                                        <td>byte/varint</td>
+                                                        <td>
+                                                            Set to <strong>value 0</strong> to denote a <i>WOTS Address</i>.
+                                                        </td>
+                                                    </tr>
+                                                    <tr>
+                                                        <td>Address</td>
+                                                        <td>ByteArray[49]</td>
+                                                        <td>The T5B1 encoded WOTS address.</td>
+                                                    </tr>
+                                                </table>
+                                            </details>
+                                            <details>
+                                                <summary>Ed25519 Address</summary>
+                                                <table>
+                                                    <tr>
+                                                        <td><b>Name<b></td>
+                                                        <td><b>Type</b></td>
+                                                        <td><b>Description</b></td>
+                                                    </tr>
+                                                    <tr>
+                                                        <td>Address Type</td>
+                                                        <td>byte/varint</td>
+                                                        <td>
+                                                            Set to <strong>value 1</strong> to denote an <i>Ed25519 Address</i>.
+                                                        </td>
+                                                    </tr>
+                                                    <tr>
+                                                        <td>Address</td>
+                                                        <td>ByteArray[32]</td>
+                                                        <td>The raw bytes of the Ed25519 address which is a BLAKE2b-256 hash of the Ed25519 public key.</td>
+                                                    </tr>
+                                                </table>
+                                            </details>
+                                        </td>
+                                    </tr>
+                                    <tr>
+                                        <td>Value</td>
+                                        <td>uint64</td>
+                                        <td>The output value.</td>
+                                    </tr>
+                                </table>
+                            </details>
+                        </td>
+                    </tr>
+                </table>
+            </details>
+        </td>
+    </tr>
+    <tr>
+        <td>Integrity hash</td>
+        <td>ByteArray[32]</td>
+        <td>
+        The SHA256 hash of all the previous fields.
+        </td>
+    </tr>
+</table>
+
+# Drawbacks
+
+Nodes need to support this new format.
+
+# Rationale and alternatives
+
+* Grouping the UTXO per transaction reduces the file size.
+* Using zlib for compression yields a ~55 MB (from ~115 MB) file on a local snapshot with ~2 million outputs (on Ed25519 addresses) from ~1 million transactions and 150 SEPs.
+
+Unlike the current format, this new format does no longer include:
+* Spent addresses: since this information is no longer held by nodes.
+* Seen milestones: as they can be requested via protocol messages.
+
+# Unresolved questions
+
+* Can we get rid of the spent addresses or do we still need to keep them?
+* Is all the information to startup a node from the local snapshot available with the described format?


text/0000-local-snapshot-file-format/0000-local-snapshot-file-format.md
```diff
@@ -0,0 +1,186 @@
...
+* start up from a recent milestone instead of having to synchronize from genesis
+* delete transaction data below the given milestone
+
+Current node implementations use a [local snapshot file format](https://github.com/iotaledger/iri-ls-sa-merger/tree/351020d3b5e342b6e9a41f2868575ab7ff8c251c#generating-an-export-file-from-a-localsnapshots-db) which only works with account based ledgers. For Chrysalis Phase 2 this file format has to be assimilated to support a UTXO based ledger.
+
+# Detailed design
+
+All types are serialized in little-endian and occur in the sequence of the rows defined below. Local snapshot files are compressed via zlib to further reduce size.
+
+SEP = solid entry point. `Array[T]` are prefixed with a varint denoting the length.
```
GalRogozinski, line 21: maybe add a one liner explanation of what is an SEP?

GalRogozinski (commented):
If nodes will write the data straight to a file and bypass the DB, they they will miss out on all the automatic feature goodies DB brings ( such as integrity checks and rollbacks)...

Maybe it is a good idea to mention that it is suggested to save the data to a db and extract the information from it to create the file?

text/0000-local-snapshot-file-format/0000-local-snapshot-file-format.md
karimodm, line 21:
```suggestion
SEP = solid entry point. `Array[T]` are prefixed with a varint denoting the array element count.
```

text/0000-local-snapshot-file-format/0000-local-snapshot-file-format.md
```diff
@@ -0,0 +1,186 @@
...
+This format describes version 1:
+<table>
+    <tr>
+        <th>Name</th>
+        <th>Type</th>
+        <th>Description</th>
+    </tr>
+    <tr>
+        <td>Version</td>
+        <td>byte</td>
```
karimodm, line 32: Version fields are specified as `varint`s in all other RFCs.

text/0000-local-snapshot-file-format/0000-local-snapshot-file-format.md
```diff
@@ -0,0 +1,186 @@
...
+        <td>Array[ByteArray[32]]</td>
+        <td>
+        The BLAKE2b-256 hashes of the SEP messages at the cut off point of the given milestone.
+        </td>
+    </tr>
+    <tr>
+        <td>UTXOs</td>
+        <td colspan="2">
+            <details open="true">
+                <summary>Array[UTXO]</summary>
```
karimodm, line 69: If I understand correctly this array will be preceded by a varint specifying the element count of the Array itself; while forming the file you may not know ahead of time how many UTXOs to persist, and therefore the precise offset of the UTXOs array, shifted by its count varint. Seeking and moving on that file descriptior might be quite annoying: perhaps it would be better to detach your Array definition from its respective varint count, so you can manipulate the type of this latter depending on the circumstance. In this case you are probably better off with a fixed type for the UTXOs count.

text/0000-local-snapshot-file-format/0000-local-snapshot-file-format.md
```diff
@@ -0,0 +1,186 @@
...
+                    </tr>
+                </table>
+            </details>
+        </td>
+    </tr>
+    <tr>
+        <td>Integrity hash</td>
+        <td>ByteArray[32]</td>
+        <td>
+        The SHA256 hash of all the previous fields.
```
karimodm, line 163: Why introducing another hash type instead of using `BLAKE2b-256`?

text/0000-local-snapshot-file-format/0000-local-snapshot-file-format.md
```diff
@@ -0,0 +1,186 @@
...
+                        <td>
+                        Unspent Outputs
+                        </td>
+                        <td colspan="2">
+                            <details open="true">
+                                <summary>Array[Outputs]</summary>
+                                <table>
+                                    <tr>
+                                        <td>Index</td>
+                                        <td>byte</td>
```
karimodm, line 94: This should be a `varint` to be consistent with [Inputs / Outputs Count fields](https://github.com/luca-moser/protocol-rfcs/blob/signed-tx-payload/text/0000-signed-transaction-payload/0000-signed-transaction-payload.md#serialized-layout), even if we "informally" define such count to be < 127.

text/0000-local-snapshot-file-format/0000-local-snapshot-file-format.md
Wollac: I was wondering about the same thing :thinking: 
What is actually the purpose of this?

If it's about authenticity it should be a signature, shouldn't it? 
If it is just about data integrity, I think this should be part of the second layer. Exactly like the compression.
Probably, I would just remove it and lift that functionality higher.

text/0000-local-snapshot-file-format/0000-local-snapshot-file-format.md
Wollac: I guess this needs to be revisited once, when exactly that part of RFC #18 has been finalized. `byte` definitely seems to restricting to me.

text/0000-local-snapshot-file-format/0000-local-snapshot-file-format.md
luca-moser: Yeah unfortunately I only wrote the export code after writing the RFC. I am currently writing up another format which also allows iterators for UTXOs.

text/0000-local-snapshot-file-format/0000-local-snapshot-file-format.md
luca-moser: Will be removed.

text/0000-local-snapshot-file-format/0000-local-snapshot-file-format.md
luca-moser: All array lengths will be replaced with fixed sized integers.

text/0000-local-snapshot-file-format/0000-local-snapshot-file-format.md
luca-moser: Please look at the new changes.

text/0000-local-snapshot-file-format/0000-local-snapshot-file-format.md
```diff
@@ -0,0 +1,149 @@
...
+seps<array[seps_count]>:
+	sep<array[32]>
+outputs<array[outputs_count]>:
+	transaction_hash<array[32]>
+	output_index<uint16>
+	address:
+		address_type<byte>
+		wots_address<array[49]>
+		||
+		address_tybe<byte>
```
thibault-martinez, line 61:
```suggestion
		address_type<byte>
```

text/0000-local-snapshot-file-format/0000-local-snapshot-file-format.md
```diff
@@ -0,0 +1,149 @@
...
+	diff_milestone_index<array[32]>
+	created_outputs_count<uint64>
+	created_outputs<array>:
+		transaction_hash<array[32]>
+		output_index<uint16>
+		address:
+			address_type<byte> = 0
+			wots_address<array[49]>
+			||
+			address_tybe<byte> = 1
```
thibault-martinez, line 74:
```suggestion
			address_type<byte> = 1
```

text/0000-local-snapshot-file-format/0000-local-snapshot-file-format.md
```diff
@@ -0,0 +1,149 @@
...
+		value<uint64>
+	consumed_outputs_count<uint64>
+	consumed_outputs<array>:
+		transaction_hash<array[32]>
+		output_index<uint16>
+		address:
+			address_type<byte>
+			wots_address<array[49]>
+			||
+			address_tybe<byte>
```
thibault-martinez, line 85:
```suggestion
			address_type<byte>
```

text/0000-local-snapshot-file-format/0000-local-snapshot-file-format.md
```diff
@@ -0,0 +1,149 @@
...
+	diff_milestone_index<array[32]>
+	created_outputs_count<uint64>
+	created_outputs<array>:
+		transaction_hash<array[32]>
+		output_index<uint16>
+		address:
+			address_type<byte> = 0
+			wots_address<array[49]>
+			||
+			address_tybe<byte> = 1
```
thibault-martinez, line 120:
```suggestion
			address_type<byte> = 1
```

text/0000-local-snapshot-file-format/0000-local-snapshot-file-format.md
```diff
@@ -0,0 +1,149 @@
...
+		value<uint64>
+	consumed_outputs_count<uint64>
+	consumed_outputs<array>:
+		transaction_hash<array[32]>
+		output_index<uint16>
+		address:
+			address_type<byte> = 0
+			wots_address<array[49]>
+			||
+			address_tybe<byte> = 1
```
thibault-martinez, line 131:
```suggestion
			address_type<byte> = 1
```

thibault-martinez (changes requested):

text/0000-local-snapshot-file-format/0000-local-snapshot-file-format.md
```diff
@@ -0,0 +1,149 @@
...
+```
+
+#### Delta Ledger State
+
+A delta ledger state local snapshot only contains the `diffs` of milestones starting from a given `ledger_milestone_index`. A node consuming such data must know the state of the ledger at `ledger_milestone_index`.
+
+![](https://i.imgur.com/bt5BUpe.png)
+
+A delta ledger state local snapshot is denoted by the type byte `1`:
+
```
GalRogozinski, line 97: Even though I agree a node should keep track of the diff at every milestone, should the snapshot file keep track of it?
I would just expect a diff from the base

text/0000-local-snapshot-file-format/0000-local-snapshot-file-format.md
```diff
@@ -0,0 +1,149 @@
...
+			wots_address<array[49]>
+			||
+			address_type<byte> = 1
+			ed25519_address<array[32]>
+		value<uint64>
+```
+
+# Drawbacks
+
+Nodes need to support this new format.
```
GalRogozinski, line 138: I guess you had nothing to fill in :-) ?

Maybe:
1. We have 2 different snapshot files which may be confusing
2. The size of the file will grow as the number of outputs increase

GalRogozinski (changes requested):
Few points/questions besides the comments:

- I am not sure I actually understand your ideas completely
In my imagination there should be one static snapshot file that represents the global snapshot and another one that just has the diff from the global that shouldn't change. I understand that you want people to share both files?
- I suppose the node will only generate the diff file upon some api calls, or every interval according to configuration? Because it probably will be better to usually update data in db that takes care of edge cases for data dumps?
- Maybe it is time we start thinking about an authentication mechanism for shared snapshot files. Something that always bothered me is that people can just start spreading cooked snapshot files around. Now that we are redefining the snapshot file maybe it is a good idea to think of some solution! Maybe add a signature field to it? And the node can verify it against a configurable public key? (just the simplest idea in my mind)

luca-moser:
@GalRogozinski 

It is up to the nodes from where they get the data from for the snapshot files but they will read it from the database mainly. Have a look at the implementation in Hornet if you're interested in an implementation.

> I am not sure I actually understand your ideas completely
In my imagination there should be one static snapshot file that represents the global snapshot and another one that just has the diff from the global that shouldn't change.

That is what this RFC is describing, two files. One containing a full ledger and one only being deltas for after that state. A global snapshot and full snapshot are equivalent.

> I understand that you want people to share both files?

You need to share both files if a node needs to bootstrap for the first time, as the delta files of course do not contain all the information needed for a node to bootstrap.

> I suppose the node will only generate the diff file upon some api calls, or every interval according to configuration? Because it probably will be better to usually update data in db that takes care of edge cases for data dumps?

This is an implementation detail. Nodes will probably create these files up on request (trigger by an API call) and/or by defined interval. Please check out the Hornet code if you're interested in an implementation detail.

> Maybe it is time we start thinking about an authentication mechanism for shared snapshot files. Something that always bothered me is that people can just start spreading cooked snapshot files around. Now that we are redefining the snapshot file maybe it is a good idea to think of some solution! Maybe add a signature field to it? And the node can verify it against a configurable public key? (just the simplest idea in my mind)

A node operator should download the snapshot files from a trusted source. I believe if you really want to ensure the data is correct, you'd have to expand the mechanism to compare multiple snapshot files from different sources.

Perhaps a tool where a user defines a list of nodes can request snapshot data from them and then compare them against each other would work.

In any case, faulty snapshots will cause the nodes to crash if the inclusion Merkle proof are off.

GalRogozinski:
Thanks for clarifying

What worries me about people that can create their own full snapshot files (rather than having a static global snapshot file) is that they can be mixed up with the wrong delta files. From what we had in the past with spent-addresses this is what I expect to happen.

> In any case, faulty snapshots will cause the nodes to crash if the inclusion Merkle proof are off.

Hmm, I think you meant that this what will happen after snapshot is loaded and fresh milestones will come in, thus making the situation same as before. However, what if we add the entire milestone (it is 560 byte assuming it has 3 signatures) to the diff file? 
So node syncing can validate the milestones and proof of inclusion to make sure the ledger state is indeed correct.
If I am not mistaken this will add ~1.76 GB to the file per year. In my opinion it is a small price to pay and it is much better than the other alternatives we mentioned here.

text/0000-local-snapshot-file-format/0000-local-snapshot-file-format.md
```diff
@@ -0,0 +1,147 @@
...
+
+This RFC defines a file format for local snapshots which is compatible with Chrysalis Phase 2.
+
+# Motivation
+
+Nodes create local snapshots to produce ledger representations at a point in time of a given milestone in order to be able to:
+* start up from a recent milestone instead of having to synchronize from the genesis transaction.
+* delete old transaction data below a given milestone.
+
+Current node implementations use a [local snapshot file format](https://github.com/iotaledger/iri-ls-sa-merger/tree/351020d3b5e342b6e9a41f2868575ab7ff8c251c#generating-an-export-file-from-a-localsnapshots-db) which only works with account based ledgers. For Chrysalis Phase 2 this file format has to be assimilated to support a UTXO based ledger.
```
charlesthompson3, line 15:
```suggestion
Nodes create local snapshots to produce ledger representations at a point in time of a given milestone to be able to:
* Start up from a recent milestone instead of having to synchronize from the genesis transaction.
* Delete old transaction data below a given milestone.

Current node implementations use a [local snapshot file format](https://github.com/iotaledger/iri-ls-sa-merger/tree/351020d3b5e342b6e9a41f2868575ab7ff8c251c#generating-an-export-file-from-a-localsnapshots-db) which only works with account based ledgers. For Chrysalis Phase 2, this file format has to be assimilated to support a UTXO based ledger.
```

text/0000-local-snapshot-file-format/0000-local-snapshot-file-format.md
```diff
@@ -0,0 +1,147 @@
...
+
+# Detailed design
+
+Since a UTXO based ledger is much larger in size, this RFC proposes two formats for snapshot files: 
+* a `full` format which represents a complete ledger state.
+* a `delta` format which only contains diffs (consumed and spent outputs) of milestones from a given milestone index onwards.
+
+This separation allows nodes to swiftly create new delta snapshot files, which then can be distributed with a companion full snapshot file to reconstruct a recent state.
+
+Unlike the current format, this new formats do not longer include spent addresses since this information is no longer held by nodes.
```
charlesthompson3, line 25:
```suggestion
# Detailed design

Since a UTXO based ledger is much larger in size, this RFC proposes two formats for snapshot files: 
* A `full` format which represents a complete ledger state.
* A `delta` format which only contains diffs (consumed and spent outputs) of milestones from a given milestone index onwards.

This separation allows nodes to swiftly create new delta snapshot files, which then can be distributed with a companion full snapshot file to reconstruct a recent state.

Unlike the current format, these new formats do not include spent addresses since this information is no longer held by nodes.
```

text/0000-local-snapshot-file-format/0000-local-snapshot-file-format.md
```diff
@@ -0,0 +1,147 @@
...
+
+#### Full Ledger State
+
+A full ledger snapshot file contains the UTXOs (`outputs` section) of a node's latest solid milestone (`ledger_milestone_index`). The `diffs` contain the diffs to rollback the `outputs` state in order to regain the ledger state of the snapshot milestone at (`seps_milestone_index`).
+
+![](https://i.imgur.com/e6WuufK.png)
+
+While the node producing such full ledger state snapshot could theoretically pre-compute the actual snapshot milestone state, this is deferred to the consumer of the data in order to speed up local snapshot creation.
+
+A full ledger state local snapshot is denoted by the type byte `0`:
```
charlesthompson3, line 39:
```suggestion
A full ledger snapshot file contains the UTXOs (`outputs` section) of a node's latest solid milestone (`ledger_milestone_index`). The `diffs` contain the diffs to rollback the `outputs` state to regain the ledger state of the snapshot milestone at (`seps_milestone_index`).

![](https://i.imgur.com/e6WuufK.png)

While the node producing such a full ledger state snapshot could theoretically pre-compute the actual snapshot milestone state, this is deferred to the consumer of the data to speed up local snapshot creation.

A full ledger state local snapshot is denoted by the type byte `0`:
```

charlesthompson3 (commented):
Here a few very minor edits, mostly for grammar and clarity; please let me know if you have any questions about them!


-Charles

text/0000-local-snapshot-file-format/0000-local-snapshot-file-format.md
karimodm: I think it makes sense to include information at every milestone actually: especially given the fact that it will be a good thing to introduce MS data as well (proof of inclusion specifically).

text/0000-local-snapshot-file-format/0000-local-snapshot-file-format.md
```diff
@@ -0,0 +1,147 @@
...
+* Start up from a recent milestone instead of having to synchronize from the genesis transaction.
+* Delete old transaction data below a given milestone.
+
+Current node implementations use a [local snapshot file format](https://github.com/iotaledger/iri-ls-sa-merger/tree/351020d3b5e342b6e9a41f2868575ab7ff8c251c#generating-an-export-file-from-a-localsnapshots-db) which only works with account based ledgers. For Chrysalis Phase 2, this file format has to be assimilated to support a UTXO based ledger.
+
+# Detailed design
+
+Since a UTXO based ledger is much larger in size, this RFC proposes two formats for snapshot files: 
+* A `full` format which represents a complete ledger state.
+* A `delta` format which only contains diffs (consumed and spent outputs) of milestones from a given milestone index onwards.
```
karimodm, line 21: > consumed and spent

So far the only output type that can be consumed is a TX, which implies spending those funds. I would remove the "consumed" term as it implies that are already outputs that can be consumed without spending: such as Coordicie's mana pledges.

text/0000-local-snapshot-file-format/0000-local-snapshot-file-format.md
```diff
@@ -0,0 +1,147 @@
...
+
+This separation allows nodes to swiftly create new delta snapshot files, which then can be distributed with a companion full snapshot file to reconstruct a recent state.
+
+Unlike the current format, these new formats do not include spent addresses since this information is no longer held by nodes.
```
karimodm, line 25: The reason for removing them is the discontinuing of WOTS (if we finally come to an agreement in that regard).

text/0000-local-snapshot-file-format/0000-local-snapshot-file-format.md
```diff
@@ -0,0 +1,147 @@
...
+
+> All types are serialized in little-endian
+
+#### Full Ledger State
+
+A full ledger snapshot file contains the UTXOs (`outputs` section) of a node's latest solid milestone (`ledger_milestone_index`). The `diffs` contain the diffs to rollback the `outputs` state to regain the ledger state of the snapshot milestone at (`seps_milestone_index`).
+
+![](https://i.imgur.com/e6WuufK.png)
+
+While the node producing such a full ledger state snapshot could theoretically pre-compute the actual snapshot milestone state, this is deferred to the consumer of the data to speed up local snapshot creation.
```
karimodm, line 37: I am starting to get a bit confused about the terminology: "actual snapshot milestone" == "seps_miletone_index"?

text/0000-local-snapshot-file-format/0000-local-snapshot-file-format.md
```diff
@@ -0,0 +1,147 @@
...
+	transaction_hash<array[32]>
+	output_index<uint16>
+	address:
+		address_type<byte>
+		wots_address<array[49]>
+		||
+		address_type<byte>
+		ed25519_address<array[32]>
+	value<uint64>
+diffs<array[diffs_count]>:
```
karimodm, line 63: There is no specific ordering of diffs you enforce here. You can have a diff for milestone X followed by a diff for milestone X+100. For integrity and validation sake it would be nicer to have an `array[ledger_milestone_index - seps_milestone_index]` of an array of diffs between milestone X and X+1, sorted by X, starting from `seps_milestone_index` and ending on `ledger_milestone_index`.

text/0000-local-snapshot-file-format/0000-local-snapshot-file-format.md
```diff
@@ -0,0 +1,147 @@
...
+# Rationale and alternatives
+
+* In conjunction with a companion full snapshot, a tool or node can "truncate" the data from a delta snapshot back to a single full snapshot. In that case, the `ledger_milestone_index` and `seps_milestone_index` would be the same. In the example above, given the full and delta snapshots, one could produce a new full snapshot for milestone 1350.
+* Since snapshots may include millions of UTXOs, code generating such files needs to stream data directly onto disk instead of keeping the entire representation in memory. In order to facilitate this, the count denotations for SEPs, UTXOs and diffs are at the beginning of the file. This allows code generating snapshot files to only have to seek back once after the actual count of elements is known.
+
+# Unresolved questions
+
+* Is all the information to startup a node from the local snapshot available with the described format?
+* Can we get rid of the spent addresses or do we still need to keep them?
+* Do we need to account for different types of outputs already? (we currently only have them deposit to addresses)
```
karimodm, line 147: I am afraid so... https://github.com/iotaledger/protocol-rfcs/pull/32/ already introduces a new output type that is crucial for validation.

text/0000-local-snapshot-file-format/0000-local-snapshot-file-format.md
```diff
@@ -0,0 +1,147 @@
...
+		value<uint64>
+```
+
+# Drawbacks
+
+Nodes need to support this new format.
+
+# Rationale and alternatives
+
+* In conjunction with a companion full snapshot, a tool or node can "truncate" the data from a delta snapshot back to a single full snapshot. In that case, the `ledger_milestone_index` and `seps_milestone_index` would be the same. In the example above, given the full and delta snapshots, one could produce a new full snapshot for milestone 1350.
```
karimodm, line 140: Here you mean that, after the truncation, the `seps_milestone_index` of the Delta will be == to the  `ledger_milestone_index` of the Full. Correct?

karimodm (changes requested):

luca-moser:
We need to change the snapshot file format to following:
* Include the entire milestone per diff
* The merkle root will be computed out of the output IDs instead of the message IDs
* The outputs within the snapshot file have to be ordered in the same order as they were for the merkle root computation

We switch away from message IDs, so one can compute whether the outputs within a milestone diff actually correspond to the milestone for that cone.

text/0000-local-snapshot-file-format/0000-local-snapshot-file-format.md
```diff
@@ -0,0 +1,147 @@
...
+network_id<uint64>
+seps_milestone_index<uint64>
+ledger_milestone_index<uint64>
+seps_count<uint64>
+outputs_count<uint64>
+diffs_count<uint64>
+seps<array[seps_count]>:
+	sep<array[32]>
+outputs<array[outputs_count]>:
+	transaction_hash<array[32]>
```
thibault-martinez, line 54: Name is confusing me, is this supposed to be a transaction id ? Not a message id ?

text/0000-local-snapshot-file-format/0000-local-snapshot-file-format.md
```diff
@@ -0,0 +1,215 @@
...
+                address_type<byte> = 1
+                ed25519_address<array[32]>
+                ||
+                ...
+            value<uint64>
+    consumed_outputs_count<uint64>
+    consumed_outputs<array>:
+        message_hash<array[32]>
+        transaction_hash<array[32]>
+        output_index<uint16>
```
thibault-martinez, line 111: I think this is missing the target TransactionId.

lzpap:
merged due to TIP refactor

PR #27, opened 2020-09-10
RFC: REST Node API

rufsam:
[Rendered](https://github.com/rufsam/protocol-rfcs/blob/master/text/0026-rest-api/0026-rest-api.md)


text/0026-rest-api/rest-api.yaml
```diff
@@ -0,0 +1,1275 @@
...
+        "404":
+          description: unsuccessful operation
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorResponse'
+              examples:
+                default:
+                  $ref: '#/components/examples/can-not-find-data-response-example'
+  /messages/by-hash/{hash}/-is-referenced-by-milestone:
```
wusyong, line 234: What do you think about adding reference state in the response context of the messages by default? `/messages/by-hash/{hash}/-is-referenced-by-milestone` and `/transaction-messages/is-confirmed` can be removed in this way. And with reference information being included in the message object, it gives more clear view about the status of the message in the Tangle imho.

text/0026-rest-api/rest-api.yaml
```diff
@@ -0,0 +1,1275 @@
...
+                type: array
+                items:
+                  $ref: '#/components/schemas/GetTipsResponse'
+              examples:
+                Example 1:
+                  $ref: '#/components/examples/get-tips-response-example-1'
+                Example 2:
+                  $ref: '#/components/examples/get-tips-response-example-2'
+  /messages/by-hash/{hash}:
+    get:
```
wusyong, line 68: I personally prefer to provide only either GET or POST version for each query APIs. Single query method isn't useful since multiple query can already do it imho. So how about we just leave a GET method for querying? `{hash}` can be a list of value. And since we probably want a maximum request boundry, it should be okay that GET URL length does has practical limit.

text/0026-rest-api/rest-api.yaml
rufsam: You're talking only about `GET /messages/by-hash/{hash}/-is-referenced-by-milestone` and  `POST/messages/by-hash/-is-referenced-by-milestone`. You're not talking about removing confirmation endpoints right?

text/0026-rest-api/rest-api.yaml
rufsam: Yes, will change that. Good idea.

laumair:
@rufsam Why `POST/messages/by-hash/-is-referenced-by-milestone` instead of `POST/messages/by-hash/is-referenced-by-milestone`? Notice the extra dash after `by-hash/`.

text/0026-rest-api/rest-api.yaml
rufsam: Changed. The desciptions of the endpoints are singular currently. Will try to come up with consistent plural forms; since there are some quite complex objects it's bit harder to come up with plural descriptions that makes sense.

text/0026-rest-api/rest-api.yaml
```diff
@@ -0,0 +1,895 @@
...
+          parent2: "78d546b46aec4557872139a48f66bc567687e8413578a14323548732358914a2"
+          payloadLength: 0
+          payload:
+            type: 2
+            index: "5350414d"
+            data: "SGVsbG8gd29ybGQh"
+    post-messages-by-hash-response-example-1:
+      value:
+        data:
+          "MESSAGE HASH X":
```
luca-moser, line 463: I think `data` should be an array of messages and not an object containing messages by their hash. The hash should be integrated into the message object within the array.

text/0026-rest-api/rest-api.yaml
```diff
@@ -0,0 +1,895 @@
...
+        data:
+          version: 1
+          parent1: "f532a53545103276b46876c473846d98648ee418468bce76df4868648dd73e5d"
+          parent2: "78d546b46aec4557872139a48f66bc567687e8413578a14323548732358914a2"
+          payloadLength: 0
+          payload:
+            type: 2
+            index: "5350414d"
+            data: "SGVsbG8gd29ybGQh"
+    post-messages-by-hash-response-example-1:
```
luca-moser, line 460: I believe this should be `get-`

text/0026-rest-api/rest-api.yaml
```diff
@@ -0,0 +1,895 @@
...
+              examples:
+                Example 1:
+                  $ref: '#/components/examples/get-tips-response-example-1'
+                Example 2:
+                  $ref: '#/components/examples/get-tips-response-example-2'
+  /messages/by-hash:
+    get:
+      tags:
+        - messages
+      summary: Find a message by its hash.
```
luca-moser, line 71: This summary talks in singular while the actual route can return multiple objects.

text/0026-rest-api/rest-api.yaml
```diff
@@ -0,0 +1,895 @@
...
+      operationId: get-messages-by-hash
+      parameters:
+        - in: query
+          name: hashes
+          schema:
+            type: array
+            items:
+              type: string
+          required: true
+          description: Hash of the message.
```
luca-moser, line 83: Should be plural.

text/0026-rest-api/rest-api.yaml
```diff
@@ -0,0 +1,895 @@
...
+  /messages/by-tag:
+    get:
+      tags:
+        - messages
+      summary: Find all messages that are marked by a given tag.
+      description: Find all messages that are marked by a given tag.
+      operationId: get-messages-by-tag
+      parameters:
+        - in: query
+          name: hashes
```
luca-moser, line 121: I believe this are tags not hashes. Specifically, there are no "tags" and the indexation payload uses the word "index" to substitute tags.

text/0026-rest-api/rest-api.yaml
```diff
@@ -0,0 +1,895 @@
...
+      responses:
+        "200":
+          description: successful operation
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/GetMessagesByTagResponse"
+              examples:
+                default:
+                  $ref: '#/components/examples/post-messages-by-tag-response-example-1'
```
luca-moser, line 137: `get-`

text/0026-rest-api/rest-api.yaml
```diff
@@ -0,0 +1,895 @@
...
+              schema:
+                $ref: '#/components/schemas/ErrorResponse'
+              examples:
+                default:
+                  $ref: '#/components/examples/can-not-find-data-response-example'
+  /messages/by-hash/is-referenced-by-milestone:
+    get:
+      tags:
+        - messages
+      summary: Check if a message is referenced by a milestone.
```
luca-moser, line 160: Plural.

text/0026-rest-api/rest-api.yaml
```diff
@@ -0,0 +1,895 @@
...
+      responses:
+        "200":
+          description: successful operation
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/GetIsReferencedByMilestoneResponse"
+              examples:
+                Example 1:
+                  $ref: '#/components/examples/post-is-referenced-by-milestone-response-example-1'
```
luca-moser, line 181: `get-`

text/0026-rest-api/rest-api.yaml
```diff
@@ -0,0 +1,895 @@
...
+              schema:
+                $ref: '#/components/schemas/ErrorResponse'
+              examples:
+                default:
+                  $ref: '#/components/examples/invalid-data-provided-response-example'
+  /transaction-messages/is-confirmed:
+    get:
+      tags:
+        - transaction-messages
+      summary: Check if a transaction-message is confirmed.
```
luca-moser, line 234: Plural

text/0026-rest-api/rest-api.yaml
```diff
@@ -0,0 +1,895 @@
...
+        "400":
+          description: unsuccessful operation
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorResponse'
+              examples:
+                default:
+                  $ref: '#/components/examples/invalid-data-provided-response-example'
+  /transaction-messages/is-confirmed:
```
luca-moser, line 230: This route is exactly the same as `​/messages​/by-hash​/is-referenced-by-milestone`. Are the hashes perhaps supposed to be transaction hashes instead?

luca-moser (changes requested):
* I believe we should rename `/transaction-messages` to just `/transactions`. For example, routes like `/transaction-messages/by-address` include the word `messages` for apparent reason.
* post examples still use `post-` where they should be `get-`
* Responses should return array of objects instead of objects where each sub object is keyed by the identifier of the particular entity.
* I am not sure whether the `peers` count in the `/info` response is useful information.
* The API does not describe any kind of authentication mechanism which I believe at the end we should have, perhaps something simple like a JWT.

text/0026-rest-api/rest-api.yaml
wusyong: I was thinking about general message/tx-message response in each api call could include the reference state already. So calling confirmation endpoints are not necessary in this way. Users could use one call to get every information about corresponding message.

text/0026-rest-api/rest-api.yaml
```diff
@@ -0,0 +1,895 @@
...
+              schema:
+                $ref: '#/components/schemas/ErrorResponse'
+              examples:
+                default:
+                  $ref: '#/components/examples/can-not-find-data-response-example'
+  /outputs/by-hash:
+    get:
+      tags:
+        - transaction-messages
+      summary: Find an output by its hash.
```
alexsporn, line 322: what is the output hash?

luca-moser: This should also be renamed to something like `outputs/by-id` because an output is definitely not identifier via a hash but rather the concatenation of transaction_hash+output_index

luca-moser (changes requested):

text/0026-rest-api/rest-api.yaml
```diff
@@ -0,0 +1,895 @@
...
+  /outputs/by-address:
+    get:
+      tags:
+        - transaction-messages
+      summary: Get all outputs that use a given address.
+      description: Get all outputs that use a given address.
+      operationId: get-outputs-by-address
+      parameters:
+        - in: query
+          name: outputs
```
luca-moser, line 371: The query parameter should be `addresses` since one is passing addresses and not output ids.

luca-moser (changes requested):

text/0026-rest-api/rest-api.yaml
rufsam: How would you do it in `GET /messages-by-tag` for example? Would you do something like 

```json
{
  "data": [
    {
      "abc": [
        {
          "hash": "1c6943b0487c92fd057d4d22ad844cc37ee27fe6fbe88e5ff0d20b2233f75b9d",
          "version": 1,
          "parent1": "ae07bf7ee5117a8fac99cb9969cfec17e74b3d5a575171e172d8a6e5bcbb237c",
          "parent2": "d33ec3e5a9e4d5759923f10771ce61d62627ec12464fc600dd46d58c542e5fd2",
          "payload": {
            "type": 2,
            "index": "abc",
            "data": "SGVsbG8gd29ybGQh"
          }
        }
      ]
    }
  ]
}
```

or would you do something like:

```json
{
  "data": [
    {
      "tag": "abc",
      "hash": "1c6943b0487c92fd057d4d22ad844cc37ee27fe6fbe88e5ff0d20b2233f75b9d",
      "version": 1,
      "parent1": "ae07bf7ee5117a8fac99cb9969cfec17e74b3d5a575171e172d8a6e5bcbb237c",
      "parent2": "d33ec3e5a9e4d5759923f10771ce61d62627ec12464fc600dd46d58c542e5fd2",
      "payload": {
        "type": 2,
        "index": "abc",
        "data": "SGVsbG8gd29ybGQh"
      }
    }
  ]
}
```

luca-moser: The first one, per tag, multiple messages can be in the array so the 2nd doesn't even work.

text/0026-rest-api/rest-api.yaml
rufsam: > 1. I believe we should rename `/transaction-messages` to just `/transactions`. For example, routes like `/transaction-messages/by-address` include the word `messages` for no apparent reason.
> 2. post examples still use `post-` where they should be `get-`
> 3. Responses should return array of objects instead of objects where each sub object is keyed by the identifier of the particular entity.
> 4. I am not sure whether the `peers` count in the `/info` response is useful information.
> 5. The API does not describe any kind of authentication mechanism which I believe at the end we should have, perhaps something simple like a JWT.

2) fixed
3) is being discussed here https://github.com/iotaledger/protocol-rfcs/pull/27#discussion_r488505450
4) fixed
5) in the end we should have an authentication mechanism, do you think this can be done at later stage?

Regarding 1.

Let me first explain what the idea behind it was. Maybe we can also solve https://github.com/iotaledger/protocol-rfcs/pull/27#discussion_r486779003.

Since not all messages include a signed-message payload the idea was to have `GET /messages/` endpoints that do things that all messages have in common and have separate `GET /transaction-messages/` endpoints that only do things that are relevant to transaction-messages.

> Are the hashes perhaps supposed to be transaction hashes instead?

No, `GET /transaction-messages/` endpoints consume message hashes not transaction hashes.

wusyong: It's not useful since `GET /messages` could already do these. Users are responsible for the logic to extract/filter transaction imho. If they really want to search certain transaction payloads, we should provide `GET /transactions` with parameters of `transaction-id`, `message-id` or more instead.

I suggest maybe `GET /transaction-messages` could be removed for now since it's not necessary at the moment. Let other endpoints complete all the works it can do.

rufsam: Since we did remove the /transaction-messages endpoints I will close this conversation.

text/0026-rest-api/rest-api.yaml
```diff
@@ -0,0 +1,707 @@
...
+      operationId: get-outputs-by-id
+      parameters:
+        - in: path
+          name: outputId
+          schema:
+            type: string
+          required: true
+          description: >-
+            Identifier of the output. An output is identified by the
+            concatenation of `transaction_id+output_index`.
```
karimodm, line 300: A sub-selector such as `output_index` would probably be most RESTful (and readeable) as another path parameter, so that: `/outputs/by-id/{transaction_id}/{output_index}`.

karimodm: The above also makes sense because you return these two separately as a result of other requests too, and therefore a concatenated selector like you propose is already a new _type_ on its own, harder to validate as well.

text/0026-rest-api/rest-api.yaml
```diff
@@ -0,0 +1,707 @@
...
+        '404':
+          description: unsuccessful operation
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorResponse'
+              examples:
+                default:
+                  $ref: '#/components/examples/can-not-find-data-response-example'
+  '/messages/by-index/{index}':
```
wusyong, line 114: What do you think about making `by-index` and other options parameters instead? So it will be `GET /messages?index={index}`, and vice versa.

alexsporn:
I did a PR with the API as implemented by HORNET.
See [Rendered Version](https://editor.swagger.io/?url=https://raw.githubusercontent.com/alexsporn/protocol-rfcs/master/text/0026-rest-api/rest-api.yaml)

text/0026-rest-api/rest-api.yaml
```diff
@@ -0,0 +1,927 @@
...
+
+    'get-message-by-id-data-response-example':
+      value:
+        hash: 82b1f7ba715d23874168f132eaa53cb7e9860cc84bca092b64e23b228bf1b8e2
+        version: 1
+        parent1: '0000000000000000000000000000000000000000000000000000000000000000'
+        parent2: '0000000000000000000000000000000000000000000000000000000000000000'
+        payload:
+          type: 0
+          transaction:
```
luca-moser, line 558: Should say `essence`

luca-moser (changes requested):
Other name changes:
* SignedTransaction -> Transaction
* UnsignedTransaction -> TransactionEssence
* SigLockedSingleDeposit -> SigLockedSingleOutput
* `transaction` field within a `Transaction` is now called `essence`

text/0026-rest-api/rest-api.yaml
alexsporn: I added a PR with these changes and the requested `data` and `error` structures
https://github.com/rufsam/protocol-rfcs/pull/2

alexsporn: See [https://editor.swagger.io/?url=https://raw.githubusercontent.com/alexsporn/protocol-rfcs/master/text/0026-rest-api/rest-api.yaml](Rendered Version)

text/0026-rest-api/rest-api.yaml
```diff
@@ -0,0 +1,983 @@
...
+                Normal:
+                  $ref: '#/components/examples/get-tips-response-example-normal'
+                Identical Tips:
+                  $ref: '#/components/examples/get-tips-response-example-same'
+  
+  '/api/v1/messages':
+    post:
+      tags:
+        - messages
+      summary: Submit a message to the node.
```
wusyong, line 88: If it sends only one message, this endpoint should be `/api/v1/message` imho.

alexsporn: The GET all and add/append endpoints are normally in plural form and the url to address a single resource is in singular form.

alexsporn: Some people like to do all the endpoints in plural form. If you would like that better we can change it.

text/0026-rest-api/rest-api.yaml
```diff
@@ -0,0 +1,983 @@
...
+              type: string
+            outputIndex:
+              type: integer
+            spent:
+              type: boolean
+            output:
+              anyOf:
+                - $ref: '#/components/schemas/SigLockedSingleOutput'
+
+    'GetAddressBalanceResponse':
```
wusyong, line 942: Could it also has a field telling if this address is spent?

alexsporn: A reusable address cannot be "spent". You either have outputs or not. The single outputs can be spent. To get the outputs use the other endpoint. This is just an "easy" way to get the balance of the address.

wusyong: Got it! yeah it's nice to see there's a way to get address balance without several endpoint calls.

text/0026-rest-api/rest-api.yaml
wusyong: Leave it as is then. I think making other APIs all in plural from will be a little bit confusing. I'm fine with the this as long as description mentions it's one message.

rufsam:
@alexsporn can you check following please?
1) children example does not match schema (replace messageIds)
2) `POST /messages` request schema: mark *required fields
3) in several examples and schemas "parentX" instead "parentXMessageHash" is used, please check

alexsporn:
See https://github.com/rufsam/protocol-rfcs/pull/3 with all the requested changes

alexsporn:
See https://github.com/rufsam/protocol-rfcs/pull/4 for latest changes

fijter (commented):
Quite happy with the state of the API; I wonder if the info endpoint might need some more information regarding limits (max allowed TX returned per index/address for example) and MQTT endpoint if enabled. We now have some libraries that just assume MQTT on the default port and if it can't connect it has no MQTT, but the port can be custom and there's no way of telling if it's enabled or not (unless it's part of the features array ofc).

alexsporn:
> We now have some libraries that just assume MQTT on the default port and if it can't connect it has no MQTT, but the port can be custom and there's no way of telling if it's enabled or not (unless it's part of the features array ofc). 

We explicitly exposed MQTT using WebSockets over the normal API port, so no extra port is needed

fijter:
> > We now have some libraries that just assume MQTT on the default port and if it can't connect it has no MQTT, but the port can be custom and there's no way of telling if it's enabled or not (unless it's part of the features array ofc).
> 
> We explicitly exposed MQTT using WebSockets over the normal API port, so no extra port is needed
> 
> See [gohornet/hornet#752](https://github.com/gohornet/hornet/pull/752)

True, but the libraries use native MQTT libraries and not MQTT over WS; I think we should support both.

rufsam:
> max allowed TX returned per index/address for example

The `index/address` endpoints do already have the `maxResults` field, so I don't think we should have this information in the `/api/v1/info` endpoint too. 
Regarding the MQTT port, I agree, it would be helpful for clients to look it up when they don't make use of WebSocket.

text/0026-rest-api/rest-api.yaml
```diff
@@ -0,0 +1,1442 @@
...
+        Get all outputs that use a given bech32-encoded address. If count equals
+        maxResults, then there might be more outputs available but those were
+        skipped for performance reasons. User should sweep the address to reduce
+        the amount of outputs.
+      parameters:
+        - in: path
+          name: address
+          schema:
+            type: string
+          example: iot1qx82lpavrafwkp0jclqvz4gzm7vs5g5g8rwr00gcm62s8456l5jh6aw0td8
```
luca-moser, line 462: replace with correct prefixed bech32 example address.

text/0026-rest-api/rest-api.yaml
```diff
@@ -0,0 +1,1442 @@
...
+            outputIDs:
+              type: array
+              items:
+                type: string
+    GetMilestoneByIndexResponse:
+      properties:
+        data:
+          type: object
+          properties:
+            milestoneIndex:
```
luca-moser, line 1412: As we changed this to `index` in the event channels, we should change it here too.

text/0026-rest-api/rest-api.yaml
```diff
@@ -0,0 +1,1442 @@
...
+              schema:
+                $ref: '#/components/schemas/GetPeersResponse'
+              examples:
+                default:
+                  $ref: '#/components/examples/get-peers-response-example'
+    post:
+      tags:
+        - peers
+      summary: Add a peer to the node.
+      description: 'Add a peer to the node. '
```
luca-moser, line 590:
```suggestion
      description: Add a peer to the node
```

text/0026-rest-api/rest-api.yaml
```diff
@@ -0,0 +1,1442 @@
...
+        '404':
+          description: unsuccessful operation
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/NotFoundErrorResponse'
+    delete:
+      tags:
+        - peers
+      summary: Remove a specific peer.
```
luca-moser, line 655:
```suggestion
      summary: Remove/disconnect a specific peer.
```

text/0026-rest-api/rest-api.yaml
```diff
@@ -0,0 +1,1442 @@
...
+      description: Remove a specific peer.
```
luca-moser, line 656:
```suggestion
      description: Remove/disconnect a specific peer.
```

luca-moser (changes requested):
Please extend the file with descriptions for every field within defined schemas so there is less ambiguity what a field is supposed to represent.

fijter (approved):
With the 2 points I mentioned earlier resolved (MQTT over WS is sufficient for now, if we need it we can add the port/host in the info endpoint but only when we have a valid use-case for it) I'm happy with this for initial release.

text/0026-rest-api/rest-api.yaml
```diff
@@ -0,0 +1,1625 @@
...
+      tags:
+        - tangle
+      summary: Returns two tips.
+      description: >-
+        Returns two tips that can be used to attach a message to. If the node can only provide a single tip,
+        `tip1MessageId` and `tip2MessageId` are identical. The provided tips can be considered as `non-lazy` and are therefore ideal for attaching a message.
+      responses:
+        '200':
+          description: >-
+            If there is at least on tip available and the node does consider itself
```
luca-moser, line 73:
```suggestion
            If there is at least one tip available and the node does consider itself
```

text/0026-rest-api/rest-api.yaml
```diff
@@ -0,0 +1,1625 @@
...
+      tags:
+        - messages
+      summary: Submit a message.
+      description: >-
+        Submit a message. The node takes care of missing* fields and tries to
+        build the message. On success, the message will be stored in the Tangle.
+        This endpoint will return the identifier of the message. *The
+        node will try to auto-fill the following fields in case they are
+        missing: `networkId`, `parent1MessageId`, `parent2MessageId`, `nonce`.
+        If `payload`  is missing, the message will be built without a payload.
```
luca-moser, line 103:
```suggestion
        If `payload` is missing, the message will be built without a payload.
```

text/0026-rest-api/rest-api.yaml
```diff
@@ -0,0 +1,1625 @@
...
+              sentHeartbeats: 0
+              droppedPackets: 0
+  schemas:
+
+    Message:
+      description: A message is the object nodes gossip around in the network. It always references two other messages that are known as parents. It is stored as a vertex on the tangle data structure that the nodes maintain.
+      properties:
+        networkId:
+          type: string
+          description: Network identifier. This field will signify whether this message was meant for mainnet, testnet, or a private net. It also tells what protocol rules apply to the message. It is first 8 bytes of the `BLAKE2b-256` hash of the concatenation of the network type and the protocol version string.
```
luca-moser, line 1094:
```suggestion
          description: Network identifier. This field signifies for which network the message is meant for. It also tells which protocol rules apply to the message. It is computed out of the first 8 bytes of the `BLAKE2b-256` hash of the concatenation of the network type and protocol version string.
```

text/0026-rest-api/rest-api.yaml
```diff
@@ -0,0 +1,1625 @@
...
+        parent1MessageId:
+          type: string
+          description: The Message ID of the first Message it references.
```
luca-moser, line 1097:
```suggestion
          description: The identifier of the first message this message references.
```

text/0026-rest-api/rest-api.yaml
```diff
@@ -0,0 +1,1625 @@
...
+        parent2MessageId:
+          type: string
+          description: The Message ID of the second Message it references.
```
luca-moser, line 1100:
```suggestion
          The identifier of the second message this message references.
```

text/0026-rest-api/rest-api.yaml
```diff
@@ -0,0 +1,1625 @@
...
+          description: Set to value 1 to denote a Milestone Payload.
+        index:
+          type: integer
+          description: The index of the milestone.
+        timestamp:
+          type: integer
+          description: The Unix timestamp at which the milestone was issued. The unix timestamp is specified in seconds.
+        parent1MessageId:
+          type: string
+          description: The Message ID of the first Message referenced by the milestone.
```
luca-moser, line 1230:
```suggestion
          description: The identifier of the first message this milestone references.
```

text/0026-rest-api/rest-api.yaml
```diff
@@ -0,0 +1,1625 @@
...
+        parent2MessageId:
+          type: string
+          description: The Message ID of the second Message referenced by the milestone.
```
luca-moser, line 1233:
```suggestion
          description: The identifier of the second message this milestone references.
```

text/0026-rest-api/rest-api.yaml
```diff
@@ -0,0 +1,1625 @@
...
+      properties:
+        data:
+          type: object
+          properties:
+            name:
+              type: string
+              description: The name of the node.
+            version:
+              type: string
+              description: The version of the node.
```
luca-moser, line 1365:
```suggestion
              description: The semantic version of the node.
```

text/0026-rest-api/rest-api.yaml
```diff
@@ -0,0 +1,1625 @@
...
+            isHealthy:
+              type: boolean
+              description: Tells whether the node is healthy or not.
+            networkId:
+              type: string
+              description: Tells whether the node operates on mainnet, testnet, or a private net. It also tells which protocol rules apply to the messages.
```
luca-moser, line 1371:
```suggestion
              description: Tells on which network the nodes operates on.
```

luca-moser (changes requested):
Looks good, just some minor description changes.

luca-moser (approved):

fijter (commented):
Looking good, I just wonder about the authentication for the api surrounding managing of neighbors; Maybe this should be documented as well here that this endpoint is only available when the feature is enabled in the node config and will return a 404 otherwise?

text/0026-rest-api/rest-api.yaml
```diff
@@ -0,0 +1,1777 @@
...
+            oneOf:
+              - $ref: '#/components/schemas/SigLockedSingleOutput'
+        payload:
+          oneOf:
+            - $ref: '#/components/schemas/IndexationPayload'
+          nullable: true
+
+    UTXOInput:
+      description: Describes an input which references an unspent transaction output to consume.
+      properties:
```
obany, line 1293: Needs required field e.g. `required: ["type", "transactionId", "transactionOutputIndex"]` 

This applies to pretty much all the objects, by default everything is marked as optional.

text/0026-rest-api/rest-api.yaml
```diff
@@ -0,0 +1,1777 @@
...
+
+    IndexationPayload:
+      description: The Indexation Payload to be embedded into a message.
+      properties:
+        type:
+          type: integer
+          description: Set to value 2 to denote a Indexation Payload.
+        index:
+          type: string
+          description: The indexation key to find/look up this message.
```
obany, line 1399: Indexation key size is between 1 and 64 bytes and should be UTF8

text/0026-rest-api/rest-api.yaml
```diff
@@ -0,0 +1,1777 @@
...
+              receivedHeartbeats: 0
+              sentMessages: 0
+              sentMessageRequests: 0
+              sentMilestoneRequests: 0
+              sentHeartbeats: 0
+              droppedPackets: 0
+  schemas:
+
+    Message:
+      description: A message is the object nodes gossip around in the network. It always references two other messages that are known as parents. It is stored as a vertex on the tangle data structure that the nodes maintain.
```
obany, line 1233: Maximum message size is 32Kb

text/0026-rest-api/rest-api.yaml
```diff
@@ -0,0 +1,1777 @@
...
+          description: The identifier of the second message this milestone references.
+        inclusionMerkleProof:
+          type: string
+          description: 256-bit hash based on the message IDs of all the not-ignored state-mutating transactions referenced by the milestone.
+        publicKeys:
+          type: array
+          items:
+            type: string
+          description: An array of public keys to validate the signatures. The keys must be in lexicographical order.
+        signature:
```
obany, line 1385: `signatures`

text/0026-rest-api/rest-api.yaml
```diff
@@ -0,0 +1,1777 @@
...
+    IndexationPayload:
+      description: The Indexation Payload to be embedded into a message.
+      properties:
+        type:
+          type: integer
+          description: Set to value 2 to denote a Indexation Payload.
+        index:
+          type: string
+          description: The indexation key to find/look up this message.
+        data:
```
obany, line 1400: Data is optional

text/0026-rest-api/rest-api.yaml
```diff
@@ -0,0 +1,1777 @@
...
+          type: string
+          description: The data to attach.
+
+    Peer:
```
obany, line 1404: Peer is missing `alias`

text/0026-rest-api/rest-api.yaml
```diff
@@ -0,0 +1,1777 @@
...
+              type: integer
+              description: If set, this message can be considered as a valid milestone message. This field therefore describes the milestone index of the involved milestone. A message can be considered as a valid milestone message if the milestone payload is valid and if the referenced parents in the milestone payload do match the referenced parents in the message itself. Note it's possible to have different milestone messages that all represent the same milestone.
+            ledgerInclusionState:
+              type: string
+              enum:
+                - included
+                - conflicting
+                - noTransaction
+              description: If `included`, the message contains a transaction that has been included in the ledger. If `conflicitng`, the message contains a transaction that has not been included in the ledger because it conflicts with another transaction. If the message does not contain a transaction, `ledgerInclusionState` is set to `noTransaction`.
+            conflictReason:
```
obany, line 1624: Maybe document what the values can be
https://github.com/gohornet/hornet/blob/a153e109ca6517e14d8153a1e72886f025667f28/pkg/model/storage/message_metadata.go#L26-L52

text/0026-rest-api/rest-api.yaml
```diff
@@ -0,0 +1,1777 @@
...
+      properties:
+        type:
+          type: integer
+          description: Set to value 0 to denote a SigLockedSingleOutput.
+        address:
+          anyOf:
+            - $ref: '#/components/schemas/Ed25519Address'
+        amount:
+          type: integer
+
```
obany, line 1314: Need to add `SigLockedDustAllowanceOutput` type, and add as an alternative to `SigLockedSingleOutput`

text/0026-rest-api/rest-api.yaml
```diff
@@ -0,0 +1,1777 @@
...
+              description: The identifier of the transaction.
+            outputIndex:
+              type: integer
+              description: The index of the output.
+            isSpent:
+              type: boolean
+              description: Tells if the output is spent or not.
+            output:
+              anyOf:
+                - $ref: '#/components/schemas/SigLockedSingleOutput'
```
obany, line 1689: add `SigLockedDustAllowanceOutput`

obany (changes requested):

obany (approved):

Thoralf-M:
What about the multiple parents from https://github.com/iotaledger/protocol-rfcs/pull/17? Shouldn't this already be adapted?

obany:
Maybe the multiple parent changes should be added after https://github.com/iotaledger/chrysalis-testnet/milestone/1 as none of the node implementations actually have this merged yet e.g. https://github.com/iotaledger/iota.go/pull/228

luca-moser (approved):
lgtm

obany:
Some changes not reflected in the RFC

`/api/v1/info` missing `bech32HRP`

Multi parent format

`/api/v1/tips` remove `tip1MessageId`, `tip2MessageId` and replace with `tipMessageIds`
`/api/v1/messages` remove `parent1MessageId`, `parent2MessageId` and replace with `parentMessageIds`
`/api/v1/messages/{messageId}/metadata` remove `parent1MessageId`, `parent2MessageId` and replace with `parentMessageIds`

Thoralf-M:
Address type needs to be updated

text/0026-rest-api/rest-api.yaml
```diff
@@ -2149,6 +2301,20 @@ components:
       required:
         - data
 
+    ReceiptsResponse:
+      properties:
+        data:
+          type: object
+          properties:
+            receipts:
```
obany, line 2409: The list of items in the array is an object with { receiptPayload, milestoneIndex }

see https://github.com/gohornet/hornet/blob/550487688dbd42ce335165420b5e5f8b9ac5ca25/pkg/model/utxo/receipt.go#L15-L20

text/0026-rest-api/rest-api.yaml
```diff
@@ -1780,6 +1884,53 @@ components:
...
+        - migratedAt
+        - final
+        - funds
+        - transaction
+
+    MigratedFundsEntry:
+      properties:
+        tailTransactionHash:
+          type: string
+        address:
```
obany, line 2018: The address is an object with { type, address }

see https://github.com/iotaledger/iota.go/blob/555f6dae0fe0faa9c7a3a5e90cf7561a17abaff9/migrated_funds_entry.go#L97

text/0026-rest-api/rest-api.yaml
```diff
@@ -1668,6 +1729,49 @@ components:
...
+        - type
+        - input
+        - output
+
+    TreasuryInput:
+      properties:
+        type:
+          type: integer
+          description: Set to value 1 to denote a TreasuryInput.
+        messageId:
```
obany, line 1857: Should be milestoneHash

https://github.com/iotaledger/iota.go/blob/555f6dae0fe0faa9c7a3a5e90cf7561a17abaff9/treasury_input.go#L64

text/0026-rest-api/rest-api.yaml
```diff
@@ -1290,6 +1333,24 @@ components:
...
+          receipts:
+            - migratedAt: 1000
+              final: false
+              funds:
+                - tailTransactionHash: NVDARJVAYEIFDKTBFQRLI9VLTVBWCJHRYBVKWPPGJAATBPAUTE9LPZQSH9UOQEYONICHVNRAFCXA99999
+                  address: 5eec99d6ee4ba21aa536c3364bbf2b587cb98a7f2565b75d948b10083e2143f8
+                  deposit: 100
+              transaction:
+                input:
+                  type: 1
```
obany, line 1443: Missing milestoneHash for input in example

text/0026-rest-api/rest-api.yaml
```diff
@@ -0,0 +1,2632 @@
...
+              f8bdbfb0f57ade7fbb95d31b11e2dbda9b2a35e9dc0cd3e11cb324e8a6bedc260100
+
+    get-receipts-response-example:
+      value:
+        data:
+          receipts:
+            - migratedAt: 1000
+              final: false
+              funds:
+                - tailTransactionHash: NVDARJVAYEIFDKTBFQRLI9VLTVBWCJHRYBVKWPPGJAATBPAUTE9LPZQSH9UOQEYONICHVNRAFCXA99999
```
Thoralf-M, line 1495: Should be hex encoded like `e0ce1933a900d02e0a43e2ba4e922d1a706a36a117ce1eb9ad0a74d39ee722c41b00cd227232efd76f5997e91bf8000000` for example

PR #29, opened 2020-11-05
PoW: swap steps, to have zero trits at the end.

c-mnd:
That's how it got implemented in https://github.com/iotaledger/iota.go/commit/a8ebed5f1a859326b89b8cac1faa5269af7e967a


c-mnd:
Sorry, I overlooked that it is already done in https://github.com/iotaledger/protocol-rfcs/pull/24.

PR #32, opened 2020-12-10
RFC: Dust Protection

GalRogozinski:
[Rendered](https://github.com/GalRogozinski/protocol-rfcs/blob/dust/text/0032-dust-protection/0032-dust-protection.md)

### DIFF
--- a/text/0032-dust-protection/0032-dust-protection.md
+++ b/text/0032-dust-protection/0032-dust-protection.md
@@ -76,15 +76,21 @@ An additional benefit of this rule is that it makes a mass of privacy violating 
 
 ### Validation
 
-Let `A` be the address that should hold the dust outputs' balances. Let `M` be the sum of all SigLockedDustAllowanceOutputs' values on `A`.So the total allowed dust outputs on `A` is (`M`/1 MIOTA) * 100 rounded down to the nearest integer. So 100 outputs for 1 MIOTA deposited.
+Let A be the address that should hold the dust outputs' balances. Let S be the sum of all the amounts of all unspent `SigLockedDustAllowanceOutputs` on A. Then, the maximum number of allowed dust outputs on A is S divided by 10,000 and rounded down, i.e. 100 outputs for each 1 Mi deposited.
 
-The SigLockedDustAllowanceOutput must be greater or equal to 1MI.
+The amount of a `SigLockedDustAllowanceOutput` must be at least 1 Mi. Apart from this, `SigLockedDustAllowanceOutputs` are processed identical to `SigLockedDustAllowanceOutput`. The transaction validation as defined in [Draft RFC-18](https://github.com/luca-moser/protocol-rfcs/blob/signed-tx-payload/text/0000-transaction-payload/0000-transaction-payload.md), however, needs to be adapted.
 
-If while processing a milestone a new dust UTXO is created but there is not a sufficient amount already locked in SigLockedDustAllowanceOutputs then the encompassing message should be marked as `ignored`.
+_Syntactical validation_ for `SigLockedDustAllowanceOutput`:
+- The `Address` must be unique in the set of `SigLockedDustAllowanceOutputs` in one transaction T. However, there can be one `SigLockedSingleOutput` and one `SigLockedDustAllowanceOutputs` T.
+- The `Amount` must be ≥ 1,000,000.
 
-If while processing a milestone an unspent SigLockedDustAllowanceOutput is consumed and as a result we have a violation on the number of allowed dust outputs then the encompassing message should be marked as `ignored`.
+The _semantic validation_ remains unchanged and are checked for both `SigLockedSingleOutputs` and `SigLockedDustAllowanceOutput`, but this RFC introduces one additional criterion:
 
-The checks should happen after the entire message was processed. Messages that were previously applied by the current milestone are taken into account.
+A transaction T
+  - consuming a `SigLockedDustAllowanceOutput` on address A **or**
+  - creating a dust output with address A,
+
+is only semantically valid, if, after T is booked, the number of unspent dust outputs on A does not exceed the allowed threshold of S / 10,000.
 
 
 # Drawbacks
--- a/text/0032-dust-protection/0032-dust-protection.md
+++ b/text/0032-dust-protection/0032-dust-protection.md
@@ -92,6 +92,58 @@ A transaction T
 
 is only semantically valid, if, after T is booked, the number of unspent dust outputs on A does not exceed the allowed threshold of S / 10,000.
 
+The following pseudocode demonstrates the new rules:
+
+```python
+boolean DustValidation(Transaction tx):
+    # maps return 0 for non-existent keys
+    dustAllowanceAddressesToBalances = new map()
+    dustAddressesToNumOfDustOutputs = new map()
+    
+    if tx not null:
+        for output in tx.outputs:
+            if output.type == SigLockedDustAllowanceOutput:
+                # must be at least 1 Mi
+                if output.balance < 1_000_000:
+                    return false
+                # only one allowance deposit per address per transaction
+                if (dustAllowanceAddressesToBalances[output.address] == 0):
+                    dustAllowanceAddressesToBalances[output.address] = output.amount
+                else
+                    return false
+            # count new dust outputs
+            if output.type == SigLockedSingleOutput & output.balance < 1_000_000:
+                dustAddressesToNumOfDustOutputs[output.address] += 1
+        for input in tx.inputs:
+            # load from the UTXO set
+            output = UTXO.loadUtxoForInput(input)
+            if output.amount < 1_000_000:
+               dustAddressesToNumOfDustOutputs[output.address] -= 1
+            if output.type == SigLockedDustAllowanceOutput:
+                dustAllowanceAddressesToBalances[output.address] =- output.amount
+        
+        addressesToValidate = SetUnion(dustAddressesToNumOfDustOutputs.keys, dustAddressesToNumOfDustOutputs.keys)
+        for address in addressesToValidate:
+            # load number of dust outputs directly from UTXO set
+            numOfDustOutputs = UTXO.loadOutputsOnAddress(address)
+                .filter(output -> output.amount < 1_000_000)
+                .count()
+            # update that number from tx data
+            numOfDustOutputs += dustAddressesToNumOfDustOutputs[address]
+            # load the allowance deposit sum from the UTXO set
+            dustAllowanceDeposit = UTXO.loadOutputsOnAddress(address)
+                                 .filter(output -> output.type == SigLockedDustAllowanceOutput)
+                                 .map(output -> output.amount)
+                                 .sum()
+            # update that number from tx data
+            dustAllowanceDeposit += dustAllowanceAddressesToBalances[address]
+
+            numOfAllowedDustOutputs = round_down(dustAllowanceDeposit/10_000)
+            if numOfDustDeposits > numOfAllowedDustOutputs:
+                return false
+        
+     return true              
+```
 
 # Drawbacks
 
--- a/text/0032-dust-protection/0032-dust-protection.md
+++ b/text/0032-dust-protection/0032-dust-protection.md
@@ -24,7 +24,7 @@ An additional benefit of this rule is that it makes a mass of privacy violating 
 
 *Dust output*: A transaction output that has an amount smaller than 1 Mi
 
-*SigLockedDustAllowanceOutput*: A new output type for deposits that enables an address to receive dust outputs. It can be consumed as an input like a regular `SigLockedSingleOutputs`.
+*SigLockedDustAllowanceOutput*: A new output type for deposits that enables an address to receive dust outputs. It can be consumed as an input like a regular `SigLockedSingleOutput`.
 
 <table>
   <tr>
--- a/text/0032-dust-protection/0032-dust-protection.md
+++ b/text/0032-dust-protection/0032-dust-protection.md
@@ -76,9 +76,7 @@ An additional benefit of this rule is that it makes a mass of privacy violating 
 
 ### Validation
 
-Let A be the address that should hold the dust outputs' balances. Let S be the sum of all the amounts of all unspent `SigLockedDustAllowanceOutputs` on A. Then, the maximum number of allowed dust outputs on A is S divided by 10,000 and rounded down, i.e. 100 outputs for each 1 Mi deposited.
-
-The amount of a `SigLockedDustAllowanceOutput` must be at least 1 Mi. Apart from this, `SigLockedDustAllowanceOutputs` are processed identical to `SigLockedSingleOutput`. The transaction validation as defined in [Draft RFC-18](https://github.com/luca-moser/protocol-rfcs/blob/signed-tx-payload/text/0000-transaction-payload/0000-transaction-payload.md), however, needs to be adapted.
+*Once a user created any number of `SigLockedDustAllowanceOutput` with at least 1 Mi, up to 127 dust outputs can be created on the same address*. Apart from this, `SigLockedDustAllowanceOutputs` are processed identical to `SigLockedSingleOutput`. The transaction validation as defined in [Draft RFC-18](https://github.com/luca-moser/protocol-rfcs/blob/signed-tx-payload/text/0000-transaction-payload/0000-transaction-payload.md), however, needs to be adapted.
 
 _Syntactical validation_ for `SigLockedDustAllowanceOutput`:
 - The `Address` must be unique in the set of `SigLockedDustAllowanceOutputs` in one transaction T. However, there can be one `SigLockedSingleOutput` and one `SigLockedDustAllowanceOutputs` T.
@@ -120,7 +118,3 @@ The second option is much more complicated as it introduces a completely new unl
 # Unresolved questions
 
 - An attacker can send microtransactions to an address with a `SigLockedDustAllowanceOutput` in order to fill the allowed threshold and block honest senders of microtransactions. The owner of the address can mitigate this by simply consolidating the attacker's dust and collecting it for profit. The problem is that the cost of doing PoW too often may exceed the profit made by collecting dust. Perhaps it is a good idea to limit "tiny" (1 i) dust outputs more than larger dust outputs? In the original discussion each dust class was supposed to be capped to a certain fixed amount. But perhaps we should give weights to different dust classes. So "tiny" dust will weigh more and fill the cap more quickly than larger dust?
-
-- Total cap per address so db key won't be polluted... should this be part of the RFC? This also depends on how key-value dbs will behave at the implementation level. RocksDb for example can configured to handle this well.
-
-- If we stay with the naive linear scheme, is 100 outputs per 1MI is really our choice?
--- a/text/0032-dust-protection/0032-dust-protection.md
+++ b/text/0032-dust-protection/0032-dust-protection.md
@@ -122,7 +122,3 @@ The second option is much more complicated as it introduces a completely new unl
 # Unresolved questions
 
 - An attacker can send microtransactions to an address with a `SigLockedDustAllowanceOutput` in order to fill the allowed threshold and block honest senders of microtransactions. The owner of the address can mitigate this by simply consolidating the attacker's dust and collecting it for profit. The problem is that the cost of doing PoW too often may exceed the profit made by collecting dust. Perhaps it is a good idea to limit "tiny" (1 i) dust outputs more than larger dust outputs? In the original discussion each dust class was supposed to be capped to a certain fixed amount. But perhaps we should give weights to different dust classes. So "tiny" dust will weigh more and fill the cap more quickly than larger dust?
-
-- Total cap per address so db key won't be polluted... should this be part of the RFC? This also depends on how key-value dbs will behave at the implementation level. RocksDb for example can configured to handle this well.
-
-- If we stay with the naive linear scheme, is 100 outputs per 1MI is really our choice?
--- a/text/0032-dust-protection/0032-dust-protection.md
+++ b/text/0032-dust-protection/0032-dust-protection.md
@@ -78,7 +78,7 @@ An additional benefit of this rule is that it makes a mass of privacy violating 
 
 Let A be the address that should hold the dust outputs' balances. Let S be the sum of all the amounts of all unspent `SigLockedDustAllowanceOutputs` on A. Then, the maximum number of allowed dust outputs on A is S divided by 10,000 and rounded down, i.e. 100 outputs for each 1 Mi deposited.
 
-The amount of a `SigLockedDustAllowanceOutput` must be at least 1 Mi. Apart from this, `SigLockedDustAllowanceOutputs` are processed identical to `SigLockedDustAllowanceOutput`. The transaction validation as defined in [Draft RFC-18](https://github.com/luca-moser/protocol-rfcs/blob/signed-tx-payload/text/0000-transaction-payload/0000-transaction-payload.md), however, needs to be adapted.
+The amount of a `SigLockedDustAllowanceOutput` must be at least 1 Mi. Apart from this, `SigLockedDustAllowanceOutputs` are processed identical to `SigLockedSingleOutput`. The transaction validation as defined in [Draft RFC-18](https://github.com/luca-moser/protocol-rfcs/blob/signed-tx-payload/text/0000-transaction-payload/0000-transaction-payload.md), however, needs to be adapted.
 
 _Syntactical validation_ for `SigLockedDustAllowanceOutput`:
 - The `Address` must be unique in the set of `SigLockedDustAllowanceOutputs` in one transaction T. However, there can be one `SigLockedSingleOutput` and one `SigLockedDustAllowanceOutputs` T.
--- a/text/0032-dust-protection/0032-dust-protection.md
+++ b/text/0032-dust-protection/0032-dust-protection.md
@@ -9,7 +9,7 @@ This RFC proposes a new protocol rule regarding the processing of outputs that t
 
 # Motivation
 
-A single typical `SigLockedSingleOutput` as described in [Draft RFC-18](https://github.com/luca-moser/protocol-rfcs/blob/signed-tx-payload/text/0000-transaction-payload/0000-transaction-payload.md) has a size of 42 bytes. When 1 Gi is split up into one billion outputs of 1 IOTA each, this corresponds to about 4 GB of outputs that need to be processed and stored. Since IOTA is feeless, this is far from a prohibitively expensive attack.
+A single typical `SigLockedSingleOutput` as described in [Draft RFC-18](https://github.com/luca-moser/protocol-rfcs/blob/signed-tx-payload/text/0000-transaction-payload/0000-transaction-payload.md) has a size of 42 bytes. When 1 Gi is split up into one billion outputs of 1 IOTA each, this corresponds to about 40 GB of outputs that need to be processed and stored. Since IOTA is feeless, this is far from a prohibitively expensive attack.
 
 In order to protect nodes from such attacks, one possible solution is to make accumulating dust outputs expensive. Since IOTA does not have any fees that might limit the feasibility of issuing many dust transactions, deposits pose a valid alternative to achieve a similar effect.
 
--- a/text/0032-dust-protection/0032-dust-protection.md
+++ b/text/0032-dust-protection/0032-dust-protection.md
@@ -10,7 +10,7 @@ This RFC proposes a new protocol rule regarding the processing of outputs that t
 # Motivation
 
 An attacker, or even honest users, can proliferate the UTXO set with outputs holding a tiny amount of iotas. This can cause the UTXO set to grow to a prohibitively large size. 
-Nodes may stall or crash due to the increasing amount of memory and computational resources needed.
+Nodes may freeze or crash due to the increasing amount of memory and computational resources needed.
 
 In order to protect nodes from such attacks, one possible solution is to make accumulating dust outputs expensive. Since IOTA does not have any fees that might limit the feasibility of issuing many dust transactions, deposits pose a valid alternative to achieve a similar effect.
 
@@ -78,7 +78,7 @@ An additional benefit of this rule is that it makes a mass of privacy violating 
 ### Validation
 
 Let A be the address that should hold the dust outputs' balances. Let S be the sum of all the amounts of all unspent `SigLockedDustAllowanceOutputs` on A. Then, the maximum number of allowed dust outputs on A is S divided by 100,000 and rounded down, i.e. 10 outputs for each 1 Mi deposited.
-There is also a cap of 100 dust outputs per address we shouldn't exceed. 
+However, regardless of S, the number of dust outputs must never exceed 100 per address.
 
 The amount of a `SigLockedDustAllowanceOutput` must be at least 1 Mi. Apart from this, `SigLockedDustAllowanceOutputs` are processed identical to `SigLockedSingleOutput`. The transaction validation as defined in [Draft RFC-18](https://github.com/luca-moser/protocol-rfcs/blob/signed-tx-payload/text/0000-transaction-payload/0000-transaction-payload.md), however, needs to be adapted.
 
@@ -92,7 +92,7 @@ A transaction T
   - consuming a `SigLockedDustAllowanceOutput` on address A **or**
   - creating a dust output with address A,
 
-is only semantically valid, if, after T is booked, the number of unspent dust outputs on A does not exceed the allowed threshold of min(S / 100,000, 100).
+is only semantically valid, if, after T is booked, the number of unspent dust outputs on A does not exceed the allowed threshold of min(S / 100000, 100).
 
 # Drawbacks
 
--- a/text/0032-dust-protection/0032-dust-protection.md
+++ b/text/0032-dust-protection/0032-dust-protection.md
@@ -76,7 +76,7 @@ An additional benefit of this rule is that it makes a mass of privacy violating 
 
 ### Validation
 
-*Once a user created any number of `SigLockedDustAllowanceOutput` with at least 1 Mi, up to 127 dust outputs can be created on the same address*. Apart from this, `SigLockedDustAllowanceOutputs` are processed identical to `SigLockedSingleOutput`. The transaction validation as defined in [Draft RFC-18](https://github.com/luca-moser/protocol-rfcs/blob/signed-tx-payload/text/0000-transaction-payload/0000-transaction-payload.md), however, needs to be adapted.
+*Once a user created any number of `SigLockedDustAllowanceOutput` with at least 1 Mi, up to 126 dust outputs can be created on the same address*. Apart from this, `SigLockedDustAllowanceOutputs` are processed identical to `SigLockedSingleOutput`. The transaction validation as defined in [Draft RFC-18](https://github.com/luca-moser/protocol-rfcs/blob/signed-tx-payload/text/0000-transaction-payload/0000-transaction-payload.md), however, needs to be adapted.
 
 _Syntactical validation_ for `SigLockedDustAllowanceOutput`:
 - The `Address` must be unique in the set of `SigLockedDustAllowanceOutputs` in one transaction T. However, there can be one `SigLockedSingleOutput` and one `SigLockedDustAllowanceOutputs` T.
@@ -99,6 +99,11 @@ is only semantically valid, if, after T is booked, the number of unspent dust ou
 
 The rationale for creating a special `SigLockedDustAllowanceOutput` rather than rely on the default `SigLockedSingleOutputs` is to prevent attackers from polluting arbitrary addresses that happen to hold
 a large amount of funds with dust.
+
+We limit the number of dust outputs to 126 so: 
+1. To incentivize users to sweep so they could collect more dust on the address.
+2. With 127 outputs a single message can sweep all the outputs and the `dustAllowanceDeposit`.
+3. A high number of outputs may have adverse affects on I/O performance.
 
 One may note that an attacker can deposit a dust allowance on 3rd party address outside his control and pollute that address with dust.
 From a security perspective this is better than an attacker depositing a dust allowance on addresses under his control.
--- a/text/0032-dust-protection/0032-dust-protection.md
+++ b/text/0032-dust-protection/0032-dust-protection.md
@@ -4,77 +4,77 @@
 
 
 # Summary
-
+In the UTXO model, each node in the network needs to keep track of all the currently unspent outputs. When the number of outputs gets too large, this can cause performance and memory issues.
 This RFC defines a new protocol rule regarding processing outputs that transfer a small amount of Iotas, known as dust outputs. The rule states that each dust output should be backed up by a deposit of the receiving address. Thus making it expensive to proliferate dust. Since a receiver must make a deposit, the protocol makes receiving dust an opt-in feature.
 
 # Motivation
 
 The total supply of IOTA is approximately 2.78 peta IOTA. Considering that a single typical `SigLockedSingleOutput` takes 42 bytes (just the plain output bytes without the encompassing transaction), it is easy to see how the size of the UTXO set may pose a potential problem. An attacker splitting 1 GI to single iota outputs can blow up the set size so that it won't fit into 4GB of RAM. This is far from being a prohibitively expensive attack.
 
-Thus, in order to protect nodes from such memory attacks we should make accumulating small dust outputs expensive. Other DLTs have fees that provide protection, but IOTA is feeless. So instead of fees we can have deposits. 
+In order to protect nodes from such attacks, one possible solution is to make accumulating dust outputs expensive. Since IOTA does not have any fees that might limit the feasibility of issuing many dust transactions, deposits pose a valid alternative to achieve a similar effect.
 
-When an address wants to receive micro transactions it must have a special output locked as a deposit. This deposit can't be consumed by any transaction as long as the dust outputs exist on the address. This also prevents attackers from sending dust to addresses that didn't opt into allowing dust outputs.
+When an address is supposed to receive micro transactions, it must have an unspent output of a special type as a deposit. This deposit cannot be consumed by any transaction as long as the dust outputs remain unspent.
 
-Additional benefit of this rule is that it makes a mass of privacy violating [forced address reuse attacks](https://en.bitcoin.it/wiki/Privacy#Forced_address_reuse) more expensive to carry out.
+An additional benefit of this rule is that it makes a mass of privacy violating [forced address reuse attacks](https://en.bitcoin.it/wiki/Privacy#Forced_address_reuse) more expensive to carry out.
 
 
 # Detailed design
 
 ### Definitions
 
-*Dust*: a UTXO that spends an amount smaller than 1 MI.
+*Dust output*: A transaction output that has an amount smaller than 1 Mi
 
-*SigLockedDustAllowanceOutput* - A new output type for deposits that allow an address to recieve dust outputs. It can be consumed as an input like a regular SigLockedOutput.
+*SigLockedDustAllowanceOutput*: A new output type for deposits that enables an address to receive dust outputs. It can be consumed as an input like a regular `SigLockedSingleOutputs`.
 
 <table>
-                                    <tr>
-                                        <td><b>Name</b></td>
-                                        <td><b>Type</b></td>
-                                        <td><b>Description</b></td>
-                                    </tr>
-                                    <tr>
-                                        <td>Output Type</td>
-                                        <td>uint8</td>
-                                        <td>
-                                            Set to <strong>value 1</strong> to denote a <i>SigLockedDustAllowanceOutput</i>.
-                                        </td>
-                                    </tr>
-                                    <tr>
-                                        <td valign="top">Address </td>
-                                        <td colspan="2">
-                                                                     <details>
-                                                <summary>Ed25519 Address</summary>
-                                                <table>
-                                                    <tr>
-                                                        <td><b>Name</b></td>
-                                                        <td><b>Type</b></td>
-                                                        <td><b>Description</b></td>
-                                                    </tr>
-                                                    <tr>
-                                                        <td>Address Type</td>
-                                                        <td>uint8</td>
-                                                        <td>
-                                                            Set to <strong>value 1</strong> to denote an <i>Ed25519 Address</i>.
-                                                        </td>
-                                                    </tr>
-                                                    <tr>
-                                                        <td>Address</td>
-                                                        <td>Array&lt;byte&gt;[32]</td>
-                                                        <td>The raw bytes of the Ed25519 address which is a BLAKE2b-256 hash of the Ed25519 public key.</td>
-                                                    </tr>
-                                                </table>
-                                            </details>
-                                        </td>
-                                    </tr>
-                                    <tr>
-                                        <td>Amount</td>
-                                        <td>uint64</td>
-                                        <td>The amount of tokens to deposit with this <i>SigLockedSingleOutput</i> output.</td>
-                                    </tr>
-                                </table>
+  <tr>
+    <td><b>Name</b></td>
+    <td><b>Type</b></td>
+    <td><b>Description</b></td>
+  </tr>
+  <tr>
+    <td>Output Type</td>
+    <td>uint8</td>
+    <td>
+      Set to <strong>value 1</strong> to denote a <i>SigLockedDustAllowanceOutput</i>.
+    </td>
+  </tr>
+  <tr>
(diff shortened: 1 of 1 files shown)


text/0032-dust-protection/0032-dust-protection.md
```diff
@@ -0,0 +1,114 @@
...
+                                </table>
+
+
+### The new protocol dust rule
+
+Let `A` be the address that should hold the dust outputs' balances. Let `M` be the sum of all SigLockedDustAllowanceOutputs' values on `A`.So the total allowed dust outputs on `A` is (`M`/1 MIOTA) * 100 rounded down to the nearest integer. So 100 outputs for 1 MIOTA deposited.
+
+The SigLockedDustAllowanceOutput must be greater or equal to 1MI.
+
+If while processing a milestone a new dust UTXO is created but there is not a sufficient amount already locked in SigLockedDustAllowanceOutputs then the encompassing message should be marked as `ignored`.
```
karimodm, line 83: I believe `ignored` here is a reference to the whiteflag RFC, put link reference.

text/0032-dust-protection/0032-dust-protection.md
```diff
@@ -0,0 +1,114 @@
...
+                                    </tr>
+                                    <tr>
+                                        <td>Amount</td>
+                                        <td>uint64</td>
+                                        <td>The amount of tokens to deposit with this <i>SigLockedSingleOutput</i> output.</td>
+                                    </tr>
+                                </table>
+
+
+### The new protocol dust rule
```
karimodm, line 77: I think this validation rules should be rewritten more technically, even in pseudo-code; as there are some ambiguities on the order of processing, especially in the context of WhiteFlag.

text/0032-dust-protection/0032-dust-protection.md
```diff
@@ -0,0 +1,114 @@
...
+
+# Motivation
+
+The total supply of IOTA is approximately 2.78 peta IOTA. Considering that a single typical `SigLockedSingleOutput` takes 42 bytes (just the plain output bytes without the encompassing transaction), it is easy to see how the size of the UTXO set may pose a potential problem. An attacker splitting 1 GI to single iota outputs can blow up the set size so that it won't fit into 4GB of RAM. This is far from being a prohibitively expensive attack.
+
+Thus, in order to protect nodes from such memory attacks we should make accumulating small dust outputs expensive. Other DLTs have fees that provide protection, but IOTA is feeless. So instead of fees we can have deposits. 
+
+When an address wants to receive micro transactions it must have a special output locked as a deposit. This deposit can't be consumed by any transaction as long as the dust outputs exist on the address. This also prevents attackers from sending dust to addresses that didn't opt into allowing dust outputs.
```
karimodm, line 16: This partially conflicts with what stated later: that one can consume the Deposit UTXOs as long as the remaining deposit(s) cover the unspent dust UTXOs.

text/0032-dust-protection/0032-dust-protection.md
karimodm, line 83: This introduces a race condition: an attacker could always create a new dust UTXO between milestones epochs for the unlocking of the deposit UTXO to be ignored. While this will perhaps not ever be a money issue (1MIOTA is a negligible amount of money) it may represent a processing issue: some use-case may rely on this deposit and address to be invalidated for good. Therefore, I think we need to define better how dust consolidation and deposit unlock look like.

text/0032-dust-protection/0032-dust-protection.md
```diff
@@ -0,0 +1,114 @@
...
+The second solution, besides being more complicated, will relief the problem from one class of nodes but not all nodes.
+
+
+# Unresolved questions
+
+- An attacker may try to fill an address with dust to fill the total cap allowed and block honest senders that want to pay the service. The service can protect by simply consolidating the attacker dust and collecting it for profit. The problem is that if the cost of doing PoW too often may exceed the bad dust profit. Perhaps it is a good idea to limit "tiny" (1 i) dust outputs more than larger dust outputs? In the original discussion each dust class was supposed to be capped to a certain fixed amount. Perhaps we should give weights to different dust classes? So "tiny" dust will weigh more and fill the cap more quickly than larger dust?
+
+- Total cap per address so db key won't be polluted... should this be part of the RFC? This also depends on how key-value dbs will behave at the implementation level. RocksDb for example can configured to handle this well.
+
+- If we stay with the naive linear scheme, is 100 outputs per 1MI is really our choice?
```
karimodm, line 114: This is a tricky point, since the deposit/dust_outputs ratio threshold can only be decreased once in place, otherwise we'd invalidate confirmed UTXOs. So we should rather start with a really conservative (high) deposit for dust UTXOs number.

karimodm (changes requested):

text/0032-dust-protection/0032-dust-protection.md
GalRogozinski: In our discussions we considered this not a serious issue

We assumed that the receiver can always consolidate the attacker's outputs and just get free money as long as the attack continues. However, your point have to be validated with bizdev probably

text/0032-dust-protection/0032-dust-protection.md
```diff
@@ -0,0 +1,114 @@
...
+Additional benefit of this rule is that it makes a mass of privacy violating [forced address reuse attacks](https://en.bitcoin.it/wiki/Privacy#Forced_address_reuse) more expensive to carry out.
+
+
+# Detailed design
+
+### Definitions
+
+*Dust*: a UTXO that spends an amount smaller than 1 MI.
+
+*SigLockedDustAllowanceOutput* - A new output type for deposits that allow an address to recieve dust outputs. It can be consumed as an input like a regular SigLockedOutput.
```
karimodm, line 27: Who can generate this output for the address? An address that does not want to support dust could be forced into it.

text/0032-dust-protection/0032-dust-protection.md
```diff
@@ -0,0 +1,114 @@
++ Feature name: dust-protection
++ Start date: 2020-12-07
++ RFC PR: [iotaledger/protocol-rfcs#0032](https://github.com/iotaledger/protocol-rfcs/pull/0032)
+
+
+# Summary
+
```
Wollac, line 7:
```suggestion
In the UTXO model, each node in the network needs to keep track of all the currently unspent outputs. When the number of outputs gets too large, this can cause performance and memory issues.
```
Maybe add one introduction sentence

text/0032-dust-protection/0032-dust-protection.md
```diff
@@ -0,0 +1,114 @@
...
+This RFC defines a new protocol rule regarding processing outputs that transfer a small amount of Iotas, known as dust outputs. The rule states that each dust output should be backed up by a deposit of the receiving address. Thus making it expensive to proliferate dust. Since a receiver must make a deposit, the protocol makes receiving dust an opt-in feature.
```
Wollac, line 8:
```suggestion
This RFC proposes a new protocol rule regarding the processing of outputs that transfer a very small amount of IOTA, so-called dust outputs: Dust outputs are only allowed when they are backed up by a certain deposit on the receiving address. This limits the amount of dust outputs Thus making it expensive to proliferate dust. Since a receiver must make a deposit, the protocol makes receiving dust an opt-in feature.
```

text/0032-dust-protection/0032-dust-protection.md
```diff
@@ -0,0 +1,114 @@
...
+
+# Motivation
+
+The total supply of IOTA is approximately 2.78 peta IOTA. Considering that a single typical `SigLockedSingleOutput` takes 42 bytes (just the plain output bytes without the encompassing transaction), it is easy to see how the size of the UTXO set may pose a potential problem. An attacker splitting 1 GI to single iota outputs can blow up the set size so that it won't fit into 4GB of RAM. This is far from being a prohibitively expensive attack.
```
Wollac, line 12:
```suggestion
A single typical `SigLockedSingleOutput` as described in [Draft RFC-18](https://github.com/luca-moser/protocol-rfcs/blob/signed-tx-payload/text/0000-transaction-payload/0000-transaction-payload.md) has a size of 42 bytes. When 1 Gi is split up into one billion outputs of 1 IOTA each, this corresponds to about 4 GB of outputs that need to be processed and stored. Since IOTA is feeless, this is far from a prohibitively expensive attack.
```
Reference the transaction RFC, it's not primarily an issue of RAM but more DB storage and processing in general.

text/0032-dust-protection/0032-dust-protection.md
```diff
@@ -0,0 +1,114 @@
...
+
+Thus, in order to protect nodes from such memory attacks we should make accumulating small dust outputs expensive. Other DLTs have fees that provide protection, but IOTA is feeless. So instead of fees we can have deposits.
```
Wollac, line 14:
```suggestion
In order to protect nodes from such attacks, one possible solution is to make accumulating dust outputs expensive. Since IOTA does not have any fees that might limit the feasibility of issuing many dust transactions, deposits pose a valid alternative to achieve a similar effect.
```

text/0032-dust-protection/0032-dust-protection.md
Wollac, line 16:
```suggestion
When an address is supposed to receive micro transactions, it must have an unspent output of a special type as a deposit. This deposit cannot be consumed by any transaction as long as the dust outputs remain unspent.
```

text/0032-dust-protection/0032-dust-protection.md
```diff
@@ -0,0 +1,114 @@
...
+
+# Motivation
+
+The total supply of IOTA is approximately 2.78 peta IOTA. Considering that a single typical `SigLockedSingleOutput` takes 42 bytes (just the plain output bytes without the encompassing transaction), it is easy to see how the size of the UTXO set may pose a potential problem. An attacker splitting 1 GI to single iota outputs can blow up the set size so that it won't fit into 4GB of RAM. This is far from being a prohibitively expensive attack.
+
+Thus, in order to protect nodes from such memory attacks we should make accumulating small dust outputs expensive. Other DLTs have fees that provide protection, but IOTA is feeless. So instead of fees we can have deposits. 
+
+When an address wants to receive micro transactions it must have a special output locked as a deposit. This deposit can't be consumed by any transaction as long as the dust outputs exist on the address. This also prevents attackers from sending dust to addresses that didn't opt into allowing dust outputs.
+
+Additional benefit of this rule is that it makes a mass of privacy violating [forced address reuse attacks](https://en.bitcoin.it/wiki/Privacy#Forced_address_reuse) more expensive to carry out.
```
Wollac, line 18:
```suggestion
An additional benefit of this rule is that it makes a mass of privacy violating [forced address reuse attacks](https://en.bitcoin.it/wiki/Privacy#Forced_address_reuse) more expensive to carry out.
```

text/0032-dust-protection/0032-dust-protection.md
```diff
@@ -0,0 +1,114 @@
...
+
+
+# Detailed design
+
+### Definitions
+
+*Dust*: a UTXO that spends an amount smaller than 1 MI.
```
Wollac, line 25:
```suggestion
*Dust output*: A transaction output that has an amount smaller than 1 Mi
```
For the general definition Dust outputs don't necessarily need to be unspent

text/0032-dust-protection/0032-dust-protection.md
Wollac, line 27:
```suggestion
*SigLockedDustAllowanceOutput*: A new output type for deposits that enables an address to receive dust outputs. It can be consumed as an input like a regular `SigLockedSingleOutputs`.
```

text/0032-dust-protection/0032-dust-protection.md
```diff
@@ -0,0 +1,114 @@
...
+                                                </table>
+                                            </details>
+                                        </td>
+                                    </tr>
+                                    <tr>
+                                        <td>Amount</td>
+                                        <td>uint64</td>
+                                        <td>The amount of tokens to deposit with this <i>SigLockedSingleOutput</i> output.</td>
+                                    </tr>
+                                </table>
```
Wollac, line 74:
```suggestion
<table>
  <tr>
    <td><b>Name</b></td>
    <td><b>Type</b></td>
    <td><b>Description</b></td>
  </tr>
  <tr>
    <td>Output Type</td>
    <td>uint8</td>
    <td>
      Set to <strong>value 1</strong> to denote a <i>SigLockedDustAllowanceOutput</i>.
    </td>
  </tr>
  <tr>
    <td valign="top">Address</td>
    <td colspan="2">
      <details>
        <summary>Ed25519 Address</summary>
        <table>
          <tr>
            <td><b>Name</b></td>
            <td><b>Type</b></td>
            <td><b>Description</b></td>
          </tr>
          <tr>
            <td>Address Type</td>
            <td>uint8</td>
            <td>
              Set to <strong>value 1</strong> to denote an <i>Ed25519 Address</i>.
            </td>
          </tr>
          <tr>
            <td>Address</td>
            <td>Array&lt;byte&gt;[32]</td>
            <td>The raw bytes of the Ed25519 address which is a BLAKE2b-256 hash of the Ed25519 public key.</td>
          </tr>
        </table>
      </details>
    </td>
  </tr>
  <tr>
    <td>Amount</td>
    <td>uint64</td>
    <td>The amount of tokens to deposit with this <i>SigLockedSingleOutput</i> output.</td>
  </tr>
</table>
```
Beautify formatting

text/0032-dust-protection/0032-dust-protection.md
Wollac, line 77:
```suggestion
### Validation
```
Let's call this validation

text/0032-dust-protection/0032-dust-protection.md
```diff
@@ -0,0 +1,114 @@
...
+
+Let `A` be the address that should hold the dust outputs' balances. Let `M` be the sum of all SigLockedDustAllowanceOutputs' values on `A`.So the total allowed dust outputs on `A` is (`M`/1 MIOTA) * 100 rounded down to the nearest integer. So 100 outputs for 1 MIOTA deposited.
+
+The SigLockedDustAllowanceOutput must be greater or equal to 1MI.
+
+If while processing a milestone a new dust UTXO is created but there is not a sufficient amount already locked in SigLockedDustAllowanceOutputs then the encompassing message should be marked as `ignored`.
+
+If while processing a milestone an unspent SigLockedDustAllowanceOutput is consumed and as a result we have a violation on the number of allowed dust outputs then the encompassing message should be marked as `ignored`.
+
+The checks should happen after the entire message was processed. Messages that were previously applied by the current milestone are taken into account.
```
Wollac, line 87:
```suggestion
Let A be the address that should hold the dust outputs' balances. Let S be the sum of all the amounts of all unspent `SigLockedDustAllowanceOutputs` on A. Then, the maximum number of allowed dust outputs on A is S divided by 10,000 and rounded down, i.e. 100 outputs for each 1 Mi deposited.

The amount of a `SigLockedDustAllowanceOutput` must be at least 1 Mi. Apart from this, `SigLockedDustAllowanceOutputs` are processed identical to `SigLockedOutputs`. The transaction validation as defined in [Draft RFC-18](https://github.com/luca-moser/protocol-rfcs/blob/signed-tx-payload/text/0000-transaction-payload/0000-transaction-payload.md), however, needs to be adapted.

_Syntactical validation_ for `SigLockedDustAllowanceOutput`:
- The `Address` must be unique in the set of `SigLockedSingleOutputs` and `SigLockedDustAllowanceOutput`.
- The `Amount` must be ≥ 1,000,000.

The _semantic validation_ remains unchanged and are checked for both `SigLockedSingleOutputs` and `SigLockedDustAllowanceOutput`, but this RFC introduces one additional criterion:

A transaction T
  - consuming a `SigLockedDustAllowanceOutput` on address A **or**
  - creating a dust output with address A,

is only semantically valid, if, after T is booked, the number of unspent dust outputs on A must not exceed the allowed threshold of S / 10,000. When T contains more than one such output, the threshold must not be exceeded on any corresponding address.
```
Made the validation rules more technical and consistent with the way it is described in the transaction RFC

text/0032-dust-protection/0032-dust-protection.md
```diff
@@ -0,0 +1,114 @@
...
+
+
+# Drawbacks
+
+1. We can't have an address that will ever hold less than 1 MI. It will be a bit hard on testnet faucets but still workable.
```
Wollac, line 92:
```suggestion
- There can no longer be addresses holding less than 1 Mi.
```
I don't think testnet faucets would be the main issue here, this could matter for quite a few projects, so maybe just leave this example out.

text/0032-dust-protection/0032-dust-protection.md
```diff
@@ -0,0 +1,114 @@
...
+2. A service receiving micropayments may fail receiving them if it didn't consolidate dust outputs or raised the deposit for the address.
```
Wollac, line 93:
```suggestion
- A service receiving micropayments may fail receiving them, if it did not consolidate dust outputs or raised the deposit for the receiving address.
```

text/0032-dust-protection/0032-dust-protection.md
```diff
@@ -0,0 +1,114 @@
...
+
+# Rationale and alternatives
+
+Other solutions that were discussed:
```
Wollac, line 97:
```suggestion
There are potential alternatives to introducing dust deposits:
```

text/0032-dust-protection/0032-dust-protection.md
```diff
@@ -0,0 +1,114 @@
...
+
+*Burning Dust* - allow dust to live for a limited amount of time on the ledger before removing it from total supply.
```
Wollac, line 99:
```suggestion
- *Burning dust*: Allow dust outputs to exists only for a limited amount of time in the ledger before removing them.
```

text/0032-dust-protection/0032-dust-protection.md
```diff
@@ -0,0 +1,114 @@
...
+
+*Sweeping dust into merkle tree* - Instead of storing all of the dust in the UTXO set, some nodes can accumulate the dust outputs to a merkle tree root. When the dust is spent, those nodes will require a merkle proof of inclusion.
```
Wollac, line 101:
```suggestion
- *Sweeping dust into Merkle trees*: Instead of burning dust outputs after some time, they are instead compressed into a Merkle tree and only the tree root is kept. In order to spend one of these compressed outputs, the corresponding Merkle audit path needs to be supplied in addition to a regular signature.
```

I think the consideration which nodes do this sweeping or not is a bit confusing and very closely related to any permanode issues we might currently have.

text/0032-dust-protection/0032-dust-protection.md
```diff
@@ -0,0 +1,114 @@
...
+
+The first solution has some attack vectors that may burn dust that a user may have wanted to consolidate. Also solutions changing the supply can be quite controversial.
```
Wollac, line 103:
```suggestion
The first option can cause issues, when dust outputs were burned before users could consolidate them. Also, changing the supply can be controversial.
```

text/0032-dust-protection/0032-dust-protection.md
```diff
@@ -0,0 +1,114 @@
...
+
+The second solution, besides being more complicated, will relief the problem from one class of nodes but not all nodes.
```
Wollac, line 105:
```suggestion
The second option is much more complicated as it introduces a completely new unlock mechanisms and requires the nodes to store the Merkle tree roots indefinitely.
```

text/0032-dust-protection/0032-dust-protection.md
```diff
@@ -0,0 +1,114 @@
...
+
+
+# Unresolved questions
+
+- An attacker may try to fill an address with dust to fill the total cap allowed and block honest senders that want to pay the service. The service can protect by simply consolidating the attacker dust and collecting it for profit. The problem is that if the cost of doing PoW too often may exceed the bad dust profit. Perhaps it is a good idea to limit "tiny" (1 i) dust outputs more than larger dust outputs? In the original discussion each dust class was supposed to be capped to a certain fixed amount. Perhaps we should give weights to different dust classes? So "tiny" dust will weigh more and fill the cap more quickly than larger dust?
```
Wollac, line 110:
```suggestion
- An attacker can send microtransactions to an address with a `SigLockedDustAllowanceOutput` in order to fill the allowed threshold and block honest senders of microtransactions. The owner of the address can mitigate this by simply consolidating the attacker's dust and collecting it for profit. The problem is that if the cost of doing PoW too often may exceed the bad dust profit. Perhaps it is a good idea to limit "tiny" (1 i) dust outputs more than larger dust outputs? In the original discussion each dust class was supposed to be capped to a certain fixed amount. But perhaps we should give weights to different dust classes. So "tiny" dust will weigh more and fill the cap more quickly than larger dust?
```

text/0032-dust-protection/0032-dust-protection.md
Wollac: I am not even sure, whether we need explain this in the context of WhiteFlag. The transaction RFC only talks about "Semantic validation" of transactions, and semantically invalid transactions are handled via white flag. So I think it would be best to just describe the dust validation in the same way splitting up into Syntactical Validation and Semantic Validation and then refer to the transaction RFC

text/0032-dust-protection/0032-dust-protection.md
Wollac: Yes, there is some race-condition/censoring where an attacker always tries to issue a dust transaction before (wrt to the WhiteFlag order) the spend transaction of a `SigLockedDustAllowanceOutputs`. I guess this is exactly, what you guys are talking about?
I think this is an important issue and we should talk about it in the RFC. This has been one of my main concerns with the dust protection without a special output type. But I think it is much less important in this scenario: First, you can just ignore SigLockedDustAllowanceOutputs, and just move everything else without any possibilities for race-conditions (this should be sufficient for the usual "wallet" use-case). Secondly, when you are under attack and really need to spend everything, you can still "dust yourself" until the limit is reached and then issue one big transaction consuming everything at once.

text/0032-dust-protection/0032-dust-protection.md
Wollac: Is this really an issue? Since the validity of dust transaction is only evaluated when they become confirmed, it would still work as expected even though the dust state on certain address might technically be invalid.

text/0032-dust-protection/0032-dust-protection.md
Wollac: Yes, a 3rd party can create a deposit on your address. But it this really an issue? You have the option to just ignore any dust and SigLockedDustAllowanceOutput and only consider your expected transactions. This would then effectively be equivalent to forbidding foreign SigLockedDustAllowanceOutput. Or, you even have the option to siphon the dust and make (small) profit. 

Forbidding foreign SigLockedDustAllowanceOutput would be possible but also complicate the validation and introduce some strange criteria here. I think, we talked about this, but keeping everything as simple as possible seemed more desirable especially since it is not a big drawback.

However, the least we should do is mention this and our motivation.

text/0032-dust-protection/0032-dust-protection.md
karimodm: It is true that you are not invalidating the unlock condition, but just the confirmation condition; nonetheless it will undermine validation of the past, in which case you will have to strictly rely on proofs of inclusions, without being able to make up your own mind on what is valid or not.

text/0032-dust-protection/0032-dust-protection.md
karimodm: > First, you can just ignore SigLockedDustAllowanceOutputs, and just move everything else without any possibilities for race-conditions (this should be sufficient for the usual "wallet" use-case).

I don't get this first point. Are you suggesting a special WF "ordering" of dust-related transactions?

text/0032-dust-protection/0032-dust-protection.md
karimodm: > Apart from this, `SigLockedDustAllowanceOutputs` are processed identical to `SigLockedDustAllowanceOutput`. 

?

karimodm: > - The `Address` must be unique in the set of `SigLockedSingleOutputs` and `SigLockedDustAllowanceOutput`.

This does not need to be true as long as each output of type `SigLockedDustAllowanceOutput` has an amount of >= 1,000,000 .

karimodm: is only semantically valid, if, after T is booked, the number of unspent dust outputs on A ~~must~~ **does** not exceed the allowed threshold of S / 10,000.

---

When T contains more than one such ~~output~~ input / consumes more than one such output

text/0032-dust-protection/0032-dust-protection.md
karimodm: mentioned of the altered total supply needs to be introduced, as it is not clear from your suggestion were these burned UTXOs go to.

text/0032-dust-protection/0032-dust-protection.md
karimodm: I think user-side added complexity should also be mentioned: a user need a way to be notified about the compression of his dust outputs into a certain merkle root, and need a way to efficiently query some permanode to obtain the corresponding audit path, bla bla. No need to be super-specific but just underline the fact that it exposes tons of complexities.

text/0032-dust-protection/0032-dust-protection.md
karimodm: The problem is that ~~if~~ the cost of doing PoW too often may exceed the ~~bad dust~~ profit made by collecting dust.

text/0032-dust-protection/0032-dust-protection.md
Wollac: I mean, as an honest user whose address got polluted by an attacker sending a SigLockedDustAllowanceOutputs and tons of dust outputs, you can just ignore all the dust and only transfer away regular SigLockedOutputs of >= 1 Mi. When doing this there cannot be any race condition and it would be the exact same result as when an honest users could "block" SigLockedDustAllowanceOutputs.

This is true without any changes to the validation or WF ordering.

text/0032-dust-protection/0032-dust-protection.md
Wollac: integrated in my suggested change

text/0032-dust-protection/0032-dust-protection.md
Wollac: > You mean `TransactionPayload`?

What I mean is that in #18 any mention of `SigLockedSingleOutput` is supposed to be replaced with "`SigLockedSingleOutput` and/or `SigLockedDustAllowanceOutput`". But I didn't wanna go through all occurrences explicitly. Any Ideas how to makes this more clear.

Wollac: > This does not need to be true as long as each output of type `SigLockedDustAllowanceOutput` has an amount of >= 1,000,000 .

Currently #18 forbids creating a transaction with more than one output to the same address. We should try to preserve this somehow. But indeed since we now have more than one output type the question is, do we want to allow a transaction with 1 `SigLockedDustAllowanceOutput` to A and 1 `SigLockedSingleOutput` to A? Or is this forbidden?

Definitely we should forbid more than one  `SigLockedDustAllowanceOutput` to A in one transaction, just we currently do for `SigLockedSingleOutput`.

Wollac: > When T contains more than one such ~output~ input / consumes more than one such output

Not 100% sure whether I understand this remark. I made some minimal changes to the previously suggested changes, is that better now?

text/0032-dust-protection/0032-dust-protection.md
Wollac: Tried to clarify this in the suggested change

text/0032-dust-protection/0032-dust-protection.md
GalRogozinski: Not really sure it is not  issue of RAM primarily...
Sure you can always do I/O swaps so the node won't crash...

It will slow down the node considerably though..

text/0032-dust-protection/0032-dust-protection.md
GalRogozinski: btw, when you go to exchanges and crypto listings (at least the ones I checked out), they usually write `MI`

text/0032-dust-protection/0032-dust-protection.md
GalRogozinski: Not sure I understand that last sentence you added...
at least I can interpret it more than 1 way

text/0032-dust-protection/0032-dust-protection.md
karimodm: That scenario is clear: as an attacker, you cannot deny a user to spend their regular funds. What I was referring to is that, as an attacker, you can deny a user to spend its last `SigLockedDustAllowanceOutput`.

text/0032-dust-protection/0032-dust-protection.md
Wollac: OK then. I've removed that sentence, it is redundant anyways, I was just trying to make it more explicit that when a transaction contains multiple dust outputs on different addresses, the criteria must hold for all of them.

karimodm (changes requested):

text/0032-dust-protection/0032-dust-protection.md
GalRogozinski: Pseudo code added

text/0032-dust-protection/0032-dust-protection.md
GalRogozinski: This md discusses the issue at length and offers mitigations:
https://hackmd.io/@iota-protocol/S1zFaBCjP

text/0032-dust-protection/0032-dust-protection.md
GalRogozinski: > This is a tricky point, since the deposit/dust_outputs ratio threshold can only be decreased once in place
The ratio can't be decreased below the validation constraints

text/0032-dust-protection/0032-dust-protection.md
```diff
@@ -0,0 +1,179 @@
...
+
+
+### Validation
+
+Let A be the address that should hold the dust outputs' balances. Let S be the sum of all the amounts of all unspent `SigLockedDustAllowanceOutputs` on A. Then, the maximum number of allowed dust outputs on A is S divided by 10,000 and rounded down, i.e. 100 outputs for each 1 Mi deposited.
+
+The amount of a `SigLockedDustAllowanceOutput` must be at least 1 Mi. Apart from this, `SigLockedDustAllowanceOutputs` are processed identical to `SigLockedSingleOutput`. The transaction validation as defined in [Draft RFC-18](https://github.com/luca-moser/protocol-rfcs/blob/signed-tx-payload/text/0000-transaction-payload/0000-transaction-payload.md), however, needs to be adapted.
+
+_Syntactical validation_ for `SigLockedDustAllowanceOutput`:
+- The `Address` must be unique in the set of `SigLockedDustAllowanceOutputs` in one transaction T. However, there can be one `SigLockedSingleOutput` and one `SigLockedDustAllowanceOutputs` T.
```
luca-moser, line 84: What's the motivation behind allowing a `SigLockedSingleOuput` + `SigLockedDustAllowanceOutput` within the same transaction targeting the same address?

GalRogozinski: I think it makes more sense that the validation rules will be enforced per output type.

It can also be useful if someone wishes to reuse and address. Imagine someone having funds in an address. Now we want to split the funds so part of them will be the allowance without moving anything out of the address.

What would be the benefits of changing this rule?

text/0032-dust-protection/0032-dust-protection.md
```diff
@@ -0,0 +1,179 @@
...
+
+The second option is much more complicated as it introduces a completely new unlock mechanisms and requires the nodes to store the Merkle tree roots indefinitely.
+
+
+# Unresolved questions
+
+- An attacker can send microtransactions to an address with a `SigLockedDustAllowanceOutput` in order to fill the allowed threshold and block honest senders of microtransactions. The owner of the address can mitigate this by simply consolidating the attacker's dust and collecting it for profit. The problem is that the cost of doing PoW too often may exceed the profit made by collecting dust. Perhaps it is a good idea to limit "tiny" (1 i) dust outputs more than larger dust outputs? In the original discussion each dust class was supposed to be capped to a certain fixed amount. But perhaps we should give weights to different dust classes. So "tiny" dust will weigh more and fill the cap more quickly than larger dust?
+
+- Total cap per address so db key won't be polluted... should this be part of the RFC? This also depends on how key-value dbs will behave at the implementation level. RocksDb for example can configured to handle this well.
```
GalRogozinski, line 177: Not part of the RFC...

We might need to paginate the db keys... we should test on implementations.

text/0032-dust-protection/0032-dust-protection.md
```diff
@@ -0,0 +1,179 @@
...
+                dustAddressesToNumOfDustOutputs[output.address] += 1
+        for input in tx.inputs:
+            # load from the UTXO set
+            output = UTXO.loadUtxoForInput(input)
+            if output.amount < 1_000_000:
+               dustAddressesToNumOfDustOutputs[output.address] -= 1
+            if output.type == SigLockedDustAllowanceOutput:
+                dustAllowanceAddressesToBalances[output.address] =- output.amount
+        
+        addressesToValidate = SetUnion(dustAddressesToNumOfDustOutputs.keys, dustAddressesToNumOfDustOutputs.keys)
```
karimodm, line 125: These two sets are the same?

text/0032-dust-protection/0032-dust-protection.md
```diff
@@ -0,0 +1,179 @@
...
+    dustAddressesToNumOfDustOutputs = new map()
+    
+    if tx not null:
+        for output in tx.outputs:
+            if output.type == SigLockedDustAllowanceOutput:
+                # must be at least 1 Mi
+                if output.balance < 1_000_000:
+                    return false
+                # only one allowance deposit per address per transaction
+                if (dustAllowanceAddressesToBalances[output.address] == 0):
```
karimodm, line 110: Here you are treating the map as initially empty...

text/0032-dust-protection/0032-dust-protection.md
```diff
@@ -0,0 +1,179 @@
...
+            # count new dust outputs
+            if output.type == SigLockedSingleOutput & output.balance < 1_000_000:
+                dustAddressesToNumOfDustOutputs[output.address] += 1
+        for input in tx.inputs:
+            # load from the UTXO set
+            output = UTXO.loadUtxoForInput(input)
+            if output.amount < 1_000_000:
+               dustAddressesToNumOfDustOutputs[output.address] -= 1
+            if output.type == SigLockedDustAllowanceOutput:
+                dustAllowanceAddressesToBalances[output.address] =- output.amount
```
karimodm, line 123: ...but here you assume that some initial deposit could be there, consumed in the input. You may therefore ending up with a value mapped for that deposit < 0. Take the case that you have a Tx that has an input SigLockedDustAllowanceOutput UTXO of 1.1Mi and a deposit output UTXO of 1.0Mi. The value in this map for this `output.address` key will be -0.1Mi.
You probably took this into account, but I find it very confusing and error-prone, especially because on line 139 and 141 you could end up with a negative number!

karimodm (changes requested):
I find the pseudo-code really hard to follow! I think it needs to be consistently simplified.

Wollac:
> I find the pseudo-code really hard to follow! I think it needs to be consistently simplified.

I agree. I was even wondering whether it should be removed completely. During the last few iterations the Validation rules got rather precise and well-defined, so IMHO they are much easier to understand and review than the pseudo code.

Wollac:
If we really wanna keep the pseudo code, it must be simplified:
E.g
- remove `if tx not null` it is fair to assume this is always given
- remove the maps and rather write something like "if there is another `SigLockedDustAllowanceOutput` with `address` in `tx`"
- ...

text/0032-dust-protection/0032-dust-protection.md
muXxer: I told wollac yesterday about a special "lock type" to stop collecting dust. This lock type needs to be signed with the private key for the address. This way you could stop collecting dust in the first milestone, and spend your funds without a race condition in the next milestone.
We came to the conclusion that we won't add this lock type now, but I just wanted to write it down, so the idea is not lost.

GalRogozinski: Why did you come to the conclusion that this shouldn't be added?
Doesn't sound too complicated

Wollac (approved):

Thoralf-M:
```
An attacker can send microtransactions to an address with a SigLockedDustAllowanceOutput in order to fill the allowed threshold and block honest senders of microtransactions.
```
Not only that, the attacker could also make it basically impossible to move all funds from that address if more dust per dust allowance output is allowed than one can have inputs in a single transaction to collect it with the dust allowance output together
Even large dust allowance outputs should not allow more than 100 dust

GalRogozinski:
@Thoralf-M 
you have a good point

But maybe the address owner can take funds out in batches?
The idea of increasing the number of outputs so that the address would not be easily blocked for dust recieval.

For example, a service can put 3 Mi and constantly consolidate 100 outputs.

Thoralf-M:
> But maybe the address owner can take funds out in batches?

@GalRogozinski if an attacker constantly sends new transaction to this address the owner of the address can only move all funds if he does the PoW faster than the attacker or if the attacker stops spamming at some point
If you want to receive more dust on an address you should create multiple dust allowance outputs on it

Thoralf-M:
Reducing the dust per output to 100 independent of the amount of the dust allowance output also protects in cases where someone with a lot iotas (1 Ti for example) creates a dust allowance output unintentional, then an attacker can just spam 100 million dust outputs to this address

GalRogozinski:
Since we have a rule that a single message can't create more than 1 output per address, it means that an attacker will have to create 100 different messages with a signature for each. A sweeper can use 100 inputs with 1 output and 1 signature in a single message. At least from tests on my personal laptop I concluded that the attacker would need to have X20 PoW power to keep up with the sweeper.

I want to point out that the current know flaws with the proposal are discussed here https://hackmd.io/@iota-protocol/S1zFaBCjP (just a draft). In the Chrysalis protocol team we talked about implementing the dust tiers idea later on. The assumption behind the dust tiers solution is that tiny dust is less useful for business purposes than larger dust. Meaning we can be very limiting on small dust outputs, but not so much on larger dust, which will be more expensive to spend in a large scale.

> If you want to receive more dust on an address you should create multiple dust allowance outputs on it
I don't understand the difference in security between having more dust allowance outputs or just having one consolidated one

As for the last comment about someone accidentally creating a 1 Ti allowance, I guess it is possible but unlikely...
The default behavior of libs would be to create regular outputs. How often would people do it accidentally? Will it amass to 1 Ti?

The reason to allow more than 100 is to make sure that businesses that want to rely on microtransactions can actually do it w.o having their addresses blocked. From a security perspective, I agree with you.

text/0032-dust-protection/0032-dust-protection.md
```diff
@@ -0,0 +1,126 @@
...
+    <td>Amount</td>
+    <td>uint64</td>
+    <td>The amount of tokens to deposit with this <i>SigLockedDustAllowanceOutput</i> output.</td>
+  </tr>
+</table>
+
+
+### Validation
+
+Let A be the address that should hold the dust outputs' balances. Let S be the sum of all the amounts of all unspent `SigLockedDustAllowanceOutputs` on A. Then, the maximum number of allowed dust outputs on A is S divided by 10,000 and rounded down, i.e. 100 outputs for each 1 Mi deposited.
```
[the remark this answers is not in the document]
muXxer: We should add an optional query paramter to the API calls to decide if we want to get all UTXO on an address, or only all non-dust UTXO, or only all dust UTXO (to sweep them in the client).

This is needed for the following API calls:

```go
	// RouteAddressBech32Outputs is the route for getting all output IDs for an address.
	// The address must be encoded in bech32.
	// GET returns the outputIDs for all outputs of this address (optional query parameters: "include-spent").
	RouteAddressBech32Outputs = "/addresses/:address/outputs"

	// RouteAddressEd25519Outputs is the route for getting all output IDs for an ed25519 address.
	// The ed25519 address must be encoded in hex.
	// GET returns the outputIDs for all outputs of this address (optional query parameters: "include-spent").
	RouteAddressEd25519Outputs = "/addresses/ed25519/:address/outputs"
```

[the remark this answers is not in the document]
muXxer: This is kinda off-topic, and only my personal opinion, but we should add another special output type if we want to unlock an address for a colored coin.
So we don't mix the dust protection logic with the colored coin logic.
For example I may want to use colored coins as a user, but I don't want an attacker to be able to dust my address.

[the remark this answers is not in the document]
GalRogozinski: - I guess there's a bigger threat from users misusing the protocol than actual attackers. It will just mean that users will have to create the dust allowance over several messages which is not so bad.
- Since dust collecting addresses should be encouraged to sweep as soon as possible I think putting this limit is a good idea.
- Definitely we should add those queries. Was waiting for that meeting with Charlie before requesting changes for the API RFC
- Maybe, not sure what the spec for colored coins will look like yet :sweat_smile: 

@Wollac, @luca-moser , @thibault-martinez , and @karimodm 
give :+1: if you agree with those modifications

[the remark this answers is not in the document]
Wollac: - I am all for a total upper limit, that was still an open question in the RFC. However, I've no idea whether 1000 is the right parameter here.
- I am not so sure about counting all outputs the same, no matter what there actual value is. With an upper limit, the Ti accident has already been resolved, hasn't it? To me this is just needlessly complicates the creation of dust addresses, doesn't provide any attack protection and is not really easier to implement on the node-side.

text/0032-dust-protection/0032-dust-protection.md
```diff
@@ -0,0 +1,126 @@
...
++ RFC PR: [iotaledger/protocol-rfcs#0032](https://github.com/iotaledger/protocol-rfcs/pull/0032)
+
+
+# Summary
+In the UTXO model, each node in the network needs to keep track of all the currently unspent outputs. When the number of outputs gets too large, this can cause performance and memory issues.
+This RFC proposes a new protocol rule regarding the processing of outputs that transfer a very small amount of IOTA, so-called dust outputs: Dust outputs are only allowed when they are backed up by a certain deposit on the receiving address. This limits the amount of dust outputs, thus making it expensive to proliferate dust. Since a receiver must make a deposit, the protocol makes receiving dust an opt-in feature.
+
+# Motivation
+
+A single typical `SigLockedSingleOutput` as described in [Draft RFC-18](https://github.com/luca-moser/protocol-rfcs/blob/signed-tx-payload/text/0000-transaction-payload/0000-transaction-payload.md) has a size of 42 bytes. When 1 Gi is split up into one billion outputs of 1 IOTA each, this corresponds to about 4 GB of outputs that need to be processed and stored. Since IOTA is feeless, this is far from a prohibitively expensive attack.
```
Wollac, line 12:
```suggestion
A single typical `SigLockedSingleOutput` as described in [Draft RFC-18](https://github.com/luca-moser/protocol-rfcs/blob/signed-tx-payload/text/0000-transaction-payload/0000-transaction-payload.md) has a size of 42 bytes. When 1 Gi is split up into one billion outputs of 1 IOTA each, this corresponds to about 40 GB of outputs that need to be processed and stored. Since IOTA is feeless, this is far from a prohibitively expensive attack.
```
1 billion * 42 bytes ≈ 40GB

GalRogozinski: yikes, missing 0

text/0032-dust-protection/0032-dust-protection.md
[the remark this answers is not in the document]
muXxer: You are right, this way it is even better.
So leave it balance based and introduce an upper limit. :+1:

shufps:
Some suggestion from my side:

move `The Amount must be ≥ 1,000,000.` from syntactical validation to semantic validation.

Reason: Amount of dust will change over time and I can't put a static value into eg the validation of the Nano Ledger app! That shouldn be validated on node side and not on hardware-wallet side.

GalRogozinski:
Hmm,
It doesn't make sense that the amount will be less than a dust amount.
I can rephrase it this way.

Should wallets in general update themselves in case of a change? That's 
up to the wallet to decide... I still think of it as syntactic 
validation, but I guess it can be viewed as semantic

Btw, why does the hardware ledger itself has to validate anything?
Shouldn't it just sign and leave validation to the software wallet on 
the laptop that is connected to the ledger?

text/0032-dust-protection/0032-dust-protection.md
```diff
@@ -9,7 +9,8 @@ This RFC proposes a new protocol rule regarding the processing of outputs that t
 
 # Motivation
 
-A single typical `SigLockedSingleOutput` as described in [Draft RFC-18](https://github.com/luca-moser/protocol-rfcs/blob/signed-tx-payload/text/0000-transaction-payload/0000-transaction-payload.md) has a size of 42 bytes. When 1 Gi is split up into one billion outputs of 1 IOTA each, this corresponds to about 4 GB of outputs that need to be processed and stored. Since IOTA is feeless, this is far from a prohibitively expensive attack.
+An attacker, or even honest users, can proliferate the UTXO set with outputs holding a tiny amount of iotas. This can cause the UTXO set to grow to a prohibitively large size. 
+Nodes may stall or crash due to the increasing amount of memory and computational resources needed.
```
Wollac, line 13:
```suggestion
Nodes may freeze or crash due to the increasing amount of memory and computational resources needed.
```

text/0032-dust-protection/0032-dust-protection.md
```diff
@@ -76,7 +77,8 @@ An additional benefit of this rule is that it makes a mass of privacy violating
 
 ### Validation
 
-Let A be the address that should hold the dust outputs' balances. Let S be the sum of all the amounts of all unspent `SigLockedDustAllowanceOutputs` on A. Then, the maximum number of allowed dust outputs on A is S divided by 10,000 and rounded down, i.e. 100 outputs for each 1 Mi deposited.
+Let A be the address that should hold the dust outputs' balances. Let S be the sum of all the amounts of all unspent `SigLockedDustAllowanceOutputs` on A. Then, the maximum number of allowed dust outputs on A is S divided by 100,000 and rounded down, i.e. 10 outputs for each 1 Mi deposited.
+There is also a cap of 100 dust outputs per address we shouldn't exceed.
```
Wollac, line 81:
```suggestion
However, regardless of S, the number of dust outputs must never exceed 100 per address.
```

text/0032-dust-protection/0032-dust-protection.md
```diff
@@ -90,7 +92,7 @@ A transaction T
   - consuming a `SigLockedDustAllowanceOutput` on address A **or**
   - creating a dust output with address A,
 
-is only semantically valid, if, after T is booked, the number of unspent dust outputs on A does not exceed the allowed threshold of S / 10,000.
+is only semantically valid, if, after T is booked, the number of unspent dust outputs on A does not exceed the allowed threshold of min(S / 100,000, 100).
```
Wollac, line 95:
```suggestion
is only semantically valid, if, after T is booked, the number of unspent dust outputs on A does not exceed the allowed threshold of min(S / 100000, 100).
```
or `100_000` would be an option, but the comma is confusing

Wollac (changes requested):

luca-moser (approved):

text/0032-dust-protection/0032-dust-protection.md
```diff
@@ -0,0 +1,124 @@
...
+  </tr>
+</table>
+
+
+### Validation
+
+Let A be the address that should hold the dust outputs' balances. Let S be the sum of all the amounts of all unspent `SigLockedDustAllowanceOutputs` on A. Then, the maximum number of allowed dust outputs on A is S divided by 100,000 and rounded down, i.e. 10 outputs for each 1 Mi deposited.
+However, regardless of S, the number of dust outputs must never exceed 100 per address.
+
+The amount of a `SigLockedDustAllowanceOutput` must be at least 1 Mi. Apart from this, `SigLockedDustAllowanceOutputs` are processed identical to `SigLockedSingleOutput`. The transaction validation as defined in [Draft RFC-18](https://github.com/luca-moser/protocol-rfcs/blob/signed-tx-payload/text/0000-transaction-payload/0000-transaction-payload.md), however, needs to be adapted.
```
thibault-martinez, line 83: > The amount of a `SigLockedDustAllowanceOutput` must be at least 1 Mi.

This is already said right after in the syntactical validation, bit repetitive.

thibault-martinez (approved):

Wollac (approved):

PR #33, opened 2021-01-06
RFC: Node Event API

luca-moser:
[Rendered](https://playground.asyncapi.io/?load=https://raw.githubusercontent.com/luca-moser/protocol-rfcs/rfc/node-event-api/text/0033-node-event-api/0033-node-event-api.yml)


obany (commented):
Looks good to me

text/0033-node-event-api/0033-node-event-api.yml
```diff
@@ -0,0 +1,163 @@
...
+      operationId: messages
+      description: Publishes newly received messages in their serialized binary form.
+      message:
+        $ref: '#/components/messages/Message'
+  messages/referenced:
+    subscribe:
+      operationId: messagesReferenced
+      description: Publishes metadata of messages newly referenced by a milestone.
+      message:
+        $ref: '#/components/messages/MessageMetadata'
```
rufsam, line 35: Do we really need all these fields for the `messages/referenced` endpoint? Or could we only include:
- messageId
- milestoneIndex
- ledgerInclusionState
- (maybe is_solid?)

and if other information is needed, the `metadata` endpoint can be subscribed for more information?

luca-moser: I don't think it makes sense to truncate the information because then you have to use multiple channels to get to your perhaps wanted data. If you look at the other fields such as `shouldPromote`/`shouldReattach` you can see that they are optional and in fact not added if the given msg already has been referenced.

obany (commented):
Wasn't this already correct with `milestoneIndex`, or is the plan to change the code to match the spec ?

https://github.com/gohornet/hornet/blob/develop/plugins/mqtt/types.go#L12

luca-moser:
@obany the code will be changed to use `index`.

rufsam (approved):

fijter (approved):
Looks solid to me, if it provides all the endpoints needed for the most important use-cases I think this is a good spec, but pending further review from developers more in the implementation side of things.

Thoralf-M:
There is still `parent1MessageId` and `parent2MessageId`

oopsmonk:
Unable to access the spec via the link since last week: https://playground.asyncapi.io/?load=https://raw.githubusercontent.com/luca-moser/protocol-rfcs/rfc/node-event-api/text/0033-node-event-api/0033-node-event-api.yml
```
Hmm. We’re having trouble finding that site.
```

thibault-martinez:
> Unable to access the spec via the link since last week: https://playground.asyncapi.io/?load=https://raw.githubusercontent.com/luca-moser/protocol-rfcs/rfc/node-event-api/text/0033-node-event-api/0033-node-event-api.yml
> 
> ```
> Hmm. We’re having trouble finding that site.
> ```

Works fine for me.

oopsmonk:
> > Unable to access the spec via the link since last week: https://playground.asyncapi.io/?load=https://raw.githubusercontent.com/luca-moser/protocol-rfcs/rfc/node-event-api/text/0033-node-event-api/0033-node-event-api.yml
> > ```
> > Hmm. We’re having trouble finding that site.
> > ```
> 
> Works fine for me.

turn out, it's DNS_PROBE_FINISHED_NXDOMAIN on chrome, firefox doesn't give any hint on this error.

Dr-Electron:
Same for me

PR #34, opened 2021-02-18
RFC: Colored Coins

lzpap:


Thoralf-M:
Why is `ColorIOTA` needed? Can't the output just always be a `SigLockedSingleOutput`?

text/0034-colored-coins/0034-colored-coins.md
```diff
@@ -0,0 +1,359 @@
...
+  `SigLockedSingleOutput`.
+- A `SigLockedColoredOutput` with `OPCODE=MOVE` and `Color=ColorIOTA` is functionally identical to a
+  `SigLockedSingleOutput`, but the latter is preferable due to its smaller size.
+
+### Dust protection
+
+Dust protection refers to preventing the unnecessary increase of the ledger database by splitting up funds into several,
+smaller outputs. Chrysalis Phase 2 introduced a new dust protection mechanism, that defines on a protocol level when
+such small outputs can be accepted.
+Dust protection doesn't change with the introduction of colored coins, a `SigLockedColoredOutput` is considered to be
```
GalRogozinski, line 328: Really? same dust solution?

Wouldn't it be better for colored coins to force the issuer to deposit funds and lock them rather than the users?

Thoralf-M:
````
One transaction might mint several coins with unique colors. The maximum amount of different colors that can be minted in a single transaction is exactly the maximum allowed number of outputs, that is 128.
One transaction might uncolor several colored coins. The maximum amount of different colors that can be uncolored in a single transaction is exactly the maximum allowed number of outputs, that is 128.
````
For `uncolored` it should be the amount of inputs and the max amount for both is 127?

**Comment by lzpap** [comment]
> Why is `ColorIOTA` needed? Can't the output just always be a `SigLockedSingleOutput`?

@Thoralf-M 
There are two ways colored coin suport can be introduced on protocol level:
 1. Keeping `SigLockedSingleOutput` for representing IOTAs, and adding `SigLockedColoredOutput` for representing colored coins.
 2. Only having one output type, `SigLockedOutput`, that can represent both IOTA and colored coins.

Current RFC version proposes Option 1, and strictly speaking, the only time we need `ColorIOTA` is when a colored coin is uncolored, and we want to signal that this `SigLockedColoredOutput` is uncolored, if you unlock it you can only spend it as regular IOTAs. Technically, the `Opcode` can encode this information, so there is no explicit need for putting `ColorIOTA` into the output, it is mentioned in the document that this is redundant.
The benefit of having the 2 output types is that `SigLockedSingleOutput` requires only 42 bytes to store, (76 for `SigLockedColoredOutput` ) representing IOTA outputs is more space efficient, because the "color" is encoded in the output type (1 byte) rather than in the actual color attribute (33 bytes).

Alternatively, there could be only one type of output that represents both IOTAs and colored coins. The color of IOTA tokens in such outputs would have to be marked with a special color, `ColorIOTA`. Also, we could get rid of opcodes and support all colored coin operations by implicitly encoding them into the transaction validation rules.

A special color, `ColorMint` would mean that this output mints new colored coins. By looking at the balance of consumed inputs and the created outputs, nodes would be able to construct and validate the transaction balance sheet. Uncoloring/recoloring would be done by consuming a colored input, but creating `ColorIOTA`/`ColorMint` outputs (equal amount to the input side) in the transaction.

Option 2 requires more space, but unifies the handling of colored and non-colored coins.

**Comment by lzpap** [comment]
> ```
> One transaction might mint several coins with unique colors. The maximum amount of different colors that can be minted in a single transaction is exactly the maximum allowed number of outputs, that is 128.
> One transaction might uncolor several colored coins. The maximum amount of different colors that can be uncolored in a single transaction is exactly the maximum allowed number of outputs, that is 128.
> ```
> 
> For `uncolored` it should be the amount of inputs and the max amount for both is 127?

As per the [transaction payload RFC](https://github.com/luca-moser/protocol-rfcs/blob/signed-tx-payload/text/0000-transaction-payload/0000-transaction-payload.md#syntactical-validation), you are right.

File: text/0034-colored-coins/0034-colored-coins.md (hunk 1 above)
Comment (lzpap) [reply text/0034-colored-coins/0034-colored-coins.md:328]: If the use case is utility tokens or security tokens, maybe yes. IMO the current chrysalis dust protection was designed with microtransactions in mind, and not thinkning about smart contract request tokens, NFTs and other fancy stuff we can do with colored coins. These different use cases introduce new requirements and constraints. At the end of the day, colored coins and smart contracts will increase the amount of load on nodes and dust in the DB, and there is no dust solution that can prevent this from happening, only control it to a certain extent.

**Comment by lunfardo314** [comment]
Where 127 limit is coming from? This a very restrictive

**Comment by Thoralf-M** [comment]
@lzpap was I meant was not if we need `SigLockedColoredOutput` at all, but why we need `SigLockedColoredOutput` with `ColorIOTA` instead of using just `SigLockedOutput` for this case because of the reduced size. Only talking about this case `it is mentioned in the document that this is redundant`, because when it's redundant, why support something that needs more space?

**Comment by alexsporn** [comment]
Why is the color 33 bytes and not 34 bytes? The `index of the output within the transaction` is an uint16, so it uses 2 bytes. `OutputIDs` are therefore 34 bytes long.

**Comment by lzpap** [comment]
> Where the 127 limit is coming from? This a very restrictive

It is coming from the maximum number of allowed inputs/outputs in [transaction payload RFC](https://github.com/luca-moser/protocol-rfcs/blob/signed-tx-payload/text/0000-transaction-payload/0000-transaction-payload.md#syntactical-validation).

**Comment by lzpap** [comment]
> Why is the color 33 bytes and not 34 bytes? The `index of the output within the transaction` is an uint16, so it uses 2 bytes. `OutputIDs` are therefore 34 bytes long.

@alexsporn  Why is it an uint16 if the maximum index that can occur is 127? What is the point of encoding in in 2 bytes if due to the max number of outputs/inputs in a transaction 1 byte is enough?
I'm fine with encoding the color in 34 bytes, but then that also should mean the max number of outputs is raised.

**Comment by lzpap** [comment]
> @lzpap was I meant was not if we need `SigLockedColoredOutput` at all, but why we need `SigLockedColoredOutput` with `ColorIOTA` instead of using just `SigLockedOutput` for this case because of the reduced size. Only talking about this case `it is mentioned in the document that this is redundant`, because when it's redundant, why support something that needs more space?

@Thoralf-M you are right, we could restrict using `SigLockedColoredOutput` with `ColorIOTA`. However, it seems like colored coins will be an essential part of any future output type, so a little redesign is needed.
 - The balance of any output becomes a vector of {color, amount} elements.
 - `ColorIOTA` denotes the color of regular IOTAs.
 - Opcodes as written in the RFC are removed. The minting and the uncoloring instructions are implicitly encoded in the transaction validation.
 - A new color is minted by specifying `ColorMint` in the output's balance.
 - The balance sheet has to be fulfilled, the **amount** of balances of the inputs must be the same as the balance of outputs.

Examples:
 1. Minting a new color: consume input of any color, create an output with `ColorMint`. If amounts of the two sides equal, transfer is valid.
 2. Uncoloring: consume input of any color, create an output with `ColorIOTA`. If amounts of the two sides equal, transfer is valid.

Then in the future when new output types are introduced (timelock, colorlock/nftlock, etc.), they are orthogonal to colored coins.
There are ways in which we can reduce the size of transactions and outputs, for example by using a "color dictionary" attached to the tx.

File: text/0034-colored-coins/0034-colored-coins.md [hunk 2]
```diff
@@ -0,0 +1,359 @@
...
+With colored coins however, the ability to specify `MINT` operation for the creation of colors, and `UNCOLOR` operation
+to destroy such colors is absolutely necessary.
+
+Opcodes affect the 3 transaction processing stages:
+
+### Syntactical Validation With Opcodes
+
+The `Opcode` field of a `SigLockedColoredOutput` may only be:
+
+- `0` to denote `MINT` operation,
```
muXxer, line 235: Would it maybe make sense to have `MOVE` as the "default operation"?
I would set it to 0, because if there is somewhere a bug in a client lib or 3rd party lib, not setting the opcode field, maybe already colored tokes would become "worthless" because they get another color.
And `MOVE` was the only possible operation in the old IOTA :shrug:  :)

text/0034-colored-coins/0034-colored-coins.md
```diff
@@ -0,0 +1,359 @@
...
+The two sides of the transaction (inputs and outputs) define a balance sheet with positive and negative sides for
+different "colors". Consuming an input adds to the positive side, while creating a new output incurs a negative balance.
+The sum of the two side of the balance sheet must be zero.
+
+**Semantic Rules for consuming inputs (previously created outputs):**
+- If the unlocked output is a `SigLockedSingleOutput`, it adds `amount`
+  of `ColorIOTA` to the positive side.
+- If the unlocked output is a `SigLockedColorOutput`, depending on the `opcode`:
+    - `MINT` or `MOVE`: Adds `amount` `color` tokens to the positive side.
+    - `UNCOLOR`: Adds `amount` `ColorIOTA` tokens to the positive side. Previously uncolored coins become regular IOTA
```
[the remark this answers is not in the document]
lzpap: I meant if you consume an output that has `UCNOLOR` opcode, that means that it was colored, then uncolored, now you can spend it as regular IOTA. Anyway, with the new version, opcodes are removed.

text/0034-colored-coins/0034-colored-coins.md
lzpap: Good idea! @muXxer what do you think about the new [version](https://github.com/iotaledger/protocol-rfcs/pull/34/commits/24b300652244d2047e855b787e2984cc88fc922a)?

BenRoyce-Society2-com:
This is a long shot:

The timestamp of the issuing message.

1. It would have to be tamperproof, no spoofing fake dates.

2. I believe UTXO has space limitations.

Two big hurdles.

Other DLTs don't implement this but with blockchains you can lookup timestamp (eg, https://github.com/in3rsha/bitcoin-utxo-dump/issues/15 )

The reasoning for speaking up here: digital assets can prove uniqueness with a colored coin but not originality.

If there were a way to get a tamper proof timestamp in there you also get originality (important for intellectual property considerations, applicable to many use case scenarios).

The whole issue can be addressed with ISCP, but getting a timestamp native to the digital asset would be extremely efficient and provide yet more excitement over colored coins and their potential applications.

PR #38, opened 2021-11-04
TIP-18 Multi-Asset Ledger and ISC Support

lzpap:
# Description

This TIP describes new output types, furthermore updated transaction structure and validation logic.

[Rendered Version](https://github.com/lzpap/tips/blob/master/tips/TIP-0018/tip-0018.md)

# Linked RFCs
 - [Byte Cost Based Dust Protection](https://github.com/iotaledger/protocol-rfcs/pull/39)


text/0038-output-types-for-tokenization-and-sc/0038-output-types-for-tokenization-and-sc.md
```diff
@@ -0,0 +1,3416 @@
...
+[Chrysalis Part 2 Transaction Payload RFC.](https://github.com/luca-moser/protocol-rfcs/blob/signed-tx-payload/text/0000-transaction-payload/0000-transaction-payload.md)
+
+<i>Transaction Payload</i> has the following structure:
+
+### Serialized Layout
+
+A <i>Transaction Payload</i> is made up of two parts:
+1. The <i>Transaction Essence</i> part which contains the inputs, outputs and an optional embedded payload.
+2. The <i>Unlock Blocks</i> which unlock the <i>Transaction Essence</i>'s inputs. In case the unlock block contains a
+3. signature, it signs the Blake2b-256 hash of the serialized <i>Transaction Essence</i> part.
```
Thoralf-M, line 204:
```suggestion
 signature, it signs the Blake2b-256 hash of the serialized <i>Transaction Essence</i> part.
```

text/0038-output-types-for-tokenization-and-sc/0038-output-types-for-tokenization-and-sc.md
```diff
@@ -0,0 +1,3416 @@
...
+* `Unlock Block Type` must either be 0, 1, 2 or 3, denoting a `Signature Unlock Block`, a `Reference Unlock block`, an
+  `Alias Unlock Block` or an `NFT Unlock Block`.
+* `Signature Unlock Blocks` must define a `Ed25519 Signature`.
+* A `Signature Unlock Block` unlocking multiple inputs must only appear once (be unique) and be positioned at the same
+  index of the first input it unlocks. All other inputs unlocked by the same `Signature Unlock Block` must have a
+  companion `Reference Unlock Block` at the same index as the corresponding input which points to the origin
+  `Signature Unlock Block`.
+* `Reference Unlock Blocks` must specify a previous `Unlock Block` which is not of type `Reference Unlock Block`. The
+  referenced index must therefore be < the index of the `Reference Unlock Block`.
+* `Alias Unlock Blocks` must specify a previous `Unlock Block` which unlocks the alias the input is locked to. The
```
thibault-martinez, line 679: For `Alias Unlock Blocks` and `NFT Unlock Blocks`, shouldn't this mention that they should be ordered ? I know it's mentioned later but this is definitely syntactic validation.

text/0038-output-types-for-tokenization-and-sc/0038-output-types-for-tokenization-and-sc.md
```diff
@@ -0,0 +1,3416 @@
...
+The timelock can be specified as a unix timestamp or as a milestone index. When specified in both ways, both conditions
+have to pass in order for the unlock to be valid.
+
+The two time representations help to protect against the possible downtime of the Coordinator. If the Coordinator is
+down, "milestone index clock" essentially stops advancing, while "real time clock" does not. An output that specifies
+time in both clocks must satisfy both conditions (AND relation).
+
+##### Additional semantic transaction validation rules:
+- An output that has <i>Timelock Milestone Index Block</i> specified must only be consumed and unlocked in a
+  transaction, if the confirming milestone index is > than the `Milestone Index` specified in the block.
```
luca-moser, line 1143: The table says:
> The milestone index/unix starting from which the output can be consumed.

which would imply it can be consumed at a bigger equal index as the one specified in the block. Same applies to the unix timelock. Clarify whether you can consume the input at the exact same index/ts or whether it must be bigger.

luca-moser (changes requested):

text/0038-output-types-for-tokenization-and-sc/0038-output-types-for-tokenization-and-sc.md
lzpap: By ordering, do you mean the fact that they can only point to a smaller index? That constraint is defined here:
> must specify a previous `Unlock Block` ...

text/0038-output-types-for-tokenization-and-sc/0038-output-types-for-tokenization-and-sc.md
lzpap: For the expiration block we say that starting from the expiration index the sender can consume it. For consistencty, let's use the same logic gor the timelock block. So yes, you should be allowed to consume the output starting from the timelock index. I'll add the greater or equal part here.

text/0038-output-types-for-tokenization-and-sc/0038-output-types-for-tokenization-and-sc.md
```diff
@@ -0,0 +1,3416 @@
...
+                                    Amount of native tokens of the given <i>Token ID</i>.
+                                </td>
+                            </tr>
+                        </table>
+                    </details>
+                </td>
+            </tr>
+            <tr>
+                <td>Blocks Count</td>
+                <td>uint16</td>
```
muXxer, line 1687: Wouldn't uint8 be enough?

text/0038-output-types-for-tokenization-and-sc/0038-output-types-for-tokenization-and-sc.md
```diff
@@ -0,0 +1,3416 @@
...
+<table>
+    <tr>
+        <td><b>Name</b></td>
+        <td><b>Value</b></td>
+        <td><b>Description</b></td>
+    </tr>
+    <tr>
+        <td>Signature Unlock Block</td>
+        <td>0</td>
+        <td>An unlock block holding one or more signatures unlocking one or more inputs.</td>
```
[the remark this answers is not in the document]
lzpap: it should be one.

text/0038-output-types-for-tokenization-and-sc/0038-output-types-for-tokenization-and-sc.md
thibault-martinez: Later in the RFC there is `Additionally, the Alias Unlock Blocks must also be ordered to prevent circular dependencies:` and same for NFT. I am not sure if that means they should ALL be ordered or not, if yes then this rule is not enough as you can have for example 4 -> 3 and 5 -> 2, they point to lower indexes but they are not ordered.

text/0038-output-types-for-tokenization-and-sc/0038-output-types-for-tokenization-and-sc.md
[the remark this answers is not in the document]
thibault-martinez: I think there are many occurrences of this in the RFC.

text/0038-output-types-for-tokenization-and-sc/0038-output-types-for-tokenization-and-sc.md
lzpap: So it is my bad wording at the end of the day :smile_cat: 

By ordering I mean here implicit ordering due to the "must point to a lower index" criteria. You don't need to order the unlock blocks per se, they refer to inputs so you actually order inputs. By requiring pointing to a lower index, we enforce ordering of dependent unlock blocks only that may cause a deadlock. We have had the same conceptual rule btw for the refernce unlock blocks, they can only point to previous unlocks also.

thibault-martinez: Got it thanks, just wanted to be sure 👍

text/0038-output-types-for-tokenization-and-sc/0038-output-types-for-tokenization-and-sc.md
lzpap: Updated `Blocks Count` to uint8, as with the `atMostOneOfEach` rule and `Block Type` being uint8, we can't really have more than 255 feature blocks.

text/0038-output-types-for-tokenization-and-sc/0038-output-types-for-tokenization-and-sc.md
```diff
@@ -0,0 +1,2903 @@
...
+### Chain Constraint in UTXO
+
+Previously created transaction outputs are destroyed when they are consumed in a subsequent transaction as an input.
+The chain constraint makes it possible to **carry the UTXO state machine state encoded in outputs across transactions.**
+When an output with chain constraint is consumed, that transaction has to create a single subsequent output that
+carries the state forward. The **state can be updated according to the transition rules defined for the given type of
+output and its current state**. As a consequence, each such output has a unique successor, and together they form a path
+or *chain* in the graph induced by the UTXO spends. Each chain is identified by its globally unique identifier.
+
+![](https://i.imgur.com/izQ4DB1.png)
```
thibault-martinez, line 874: Would be nice if this image had a white background instead of transparent otherwise it's pretty hard to read on a dark theme (black letters on black background)

text/0038-output-types-for-tokenization-and-sc/0038-output-types-for-tokenization-and-sc.md
```diff
@@ -0,0 +1,2903 @@
...
+                                            </tr>
+                                            <tr>
+                                                <td>Address</td>
+                                                <td>ByteArray[32]</td>
+                                                <td>The raw bytes of the Ed25519 address which is a BLAKE2b-256 hash of the Ed25519 public key.</td>
+                                            </tr>
+                                        </table>
+                                    </details>
+                                    <details>
+                                        <summary>BLS Address</summary>
```
thibault-martinez, line 1215: Should be removed

text/0038-output-types-for-tokenization-and-sc/0038-output-types-for-tokenization-and-sc.md
```diff
@@ -0,0 +1,2903 @@
...
```
thibault-martinez, line 1730: Should be removed

text/0038-output-types-for-tokenization-and-sc/0038-output-types-for-tokenization-and-sc.md
lzpap: Updated :)

text/0038-output-types-for-tokenization-and-sc/0038-output-types-for-tokenization-and-sc.md
```diff
@@ -0,0 +1,2994 @@
...
+    <tr>
+        <td>Block Type</td>
+        <td>uint8</td>
+        <td>
+            Set to <strong>value 4</strong> to denote a <i>Timelock Unix Block</i>.
+        </td>
+    </tr>
+    <tr>
+        <td>Unix Time</td>
+        <td>uint32</td>
```
PhilippGackstatter, line 678:
```suggestion
        <td>uint64</td>
```

text/0038-output-types-for-tokenization-and-sc/0038-output-types-for-tokenization-and-sc.md
```diff
@@ -0,0 +1,2994 @@
...
+    <tr>
+        <td>Block Type</td>
+        <td>uint8</td>
+        <td>
+            Set to <strong>value 6</strong> to denote a <i>Expiration Unix Block</i>.
+        </td>
+    </tr>
+    <tr>
+        <td>Unix Time</td>
+        <td>uint32</td>
```
PhilippGackstatter, line 764:
```suggestion
        <td>uint64</td>
```

PhilippGackstatter (commented):
Unix timestamps should probably be `uint64` for future proofing.

text/0038-output-types-for-tokenization-and-sc/0038-output-types-for-tokenization-and-sc.md
```diff
@@ -0,0 +1,2994 @@
...
+entries in nodes' databases.
+
+##### Additional syntactic transaction validation rules:
+- An output with <i>Indexation Block</i> is valid, if and only if 0 < `Indexation Tag Length` ≤
+  `Max Indexation Tag Length`.
+
+<details>
+<summary>Indexation Block</summary>
+<blockquote>
+    Defines an indexation tag to which the <i>Metadata Block</i> will be indexed. Creates an indexation lookup in nodes.
```
thibault-martinez, line 832: Why specify `Metadata Block` ? There is not necessarily one.

lzpap: Right, it is the output that is indexed anyway. Whether it has metadata block or not doesn't matter.

text/0038-output-types-for-tokenization-and-sc/0038-output-types-for-tokenization-and-sc.md
lzpap: It has been considered already and the decision was to go with uint32.
With that, we are fine until February 7, 2106, while sparing 4 additional bytes per time value in outputs that would be zero for the next 80+ years anyway. Let's set a reminder for 2100 not to forget to update then :smile_cat:

ps: Bitcoin also uses [uint32 block timestamps](https://en.bitcoin.it/wiki/Protocol_documentation).

PhilippGackstatter: Fair enough! I stumbled upon this and thought I would point it out. I figured it would make sense to be consistent with the [milestone payload](https://github.com/iotaledger/protocol-rfcs/blob/master/text/0019-milestone-payload/0019-milestone-payload.md#structure) which uses `uint64` for the timestamp (at least in the RFC). But I guess 2106 seems fine :smile:.

text/0038-output-types-for-tokenization-and-sc/0038-output-types-for-tokenization-and-sc.md
```diff
@@ -0,0 +1,2886 @@
...
+                        <td>ByteArray[20]</td>
+                        <td>The raw bytes of the nft address which is a BLAKE2b-160 hash of the outputID that created it.</td>
+                    </tr>
+                </table>
+            </details>
+        </td>
+    </tr>
+</table>
+
+| :information_source:  Good to know about address format |
```
thibault-martinez, line 372: Maybe I'm being a bit pedantic now but I think it would be clearer to have an Address section and not wait until we first encounter addresses to explain this. Just a suggestion :)

text/0038-output-types-for-tokenization-and-sc/0038-output-types-for-tokenization-and-sc.md
```diff
@@ -0,0 +1,2886 @@
...
+
+- A token scheme is a list of hard coded constraints. It is not feasible at the moment to foresee the future
+  needs/requirements of hard coded constraints, so it is impossible to design token schemes as any possible combination
+  of those constraints. A better design would be to have a list of possible constraints (and their related fields) from
+  which the user can choose. The chosen combination should still be encoded as a bitmask inside the `Token ID`.
+- For now, only token scheme `0` is supported. Additional token schemes will be designed iteratively while doing the
+  testnet implementation.
+- The `Foundry ID` of a foundry output should have a global mapping table in nodes, so that given a `Foundry ID`, the
+  `Output ID` of the foundry output can be retrieved. `Foundry ID` behaves like an address that can't unlock anything.
+  While it is not neccessarly needed for the protocol, it is needed for client side operations (what is the current
```
thibault-martinez, line 2112:
```suggestion
  While it is not necessarily needed for the protocol, it is needed for client side operations (what is the current
```

text/0038-output-types-for-tokenization-and-sc/0038-output-types-for-tokenization-and-sc.md
```diff
@@ -0,0 +1,2886 @@
...
+        <td>NFT Reference Unlock Index</td>
+        <td>uint16</td>
+        <td>
+            Index of input and unlock block corresponding to an NFT output.
+        </td>
+    </tr>
+</table>
+
+An *Alias Unlock Block* is only valid if the input in the transaction at index `NFT Reference Unlock Index` is the NFT
+output with the same `NFT ID` as the `Address` field of the to-be unlocked output.
```
oopsmonk, line 2835: should be NFT?
```suggestion
An *NFT Unlock Block* is only valid if the input in the transaction at index `NFT Reference Unlock Index` is the NFT
output with the same `NFT ID` as the `Address` field of the to-be unlocked output.
```

tips/TIP-0018/tip-0018.md
```diff
@@ -0,0 +1,3755 @@
...
+Therefore, **every output must be able to hold not only IOTA coins, but also native tokens**.
+
+Dust protection applies to all outputs, therefore it is not possible for outputs to hold only native tokens, the dust
+requirements must be covered via IOTA coins.
+
+User defined tokens are called <i>Native Tokens</i> on protocol level. The maximum supply of a particular native token
+is defined by the representation chosen on protocol level for defining their amounts in outputs. Since native tokens
+are also a vehicle to wrap layer 2 tokens into layer 1 tokens, the chosen representation must take into account the
+maximum possible supply of layer 2 tokens. Solidity, the most popular layer 2 smart contract language defines the
+maximum supply of an ERC-20 token as `MaxUint256`, therefore it should be possible represent such huge amount of assets
```
thibault-martinez: `it should be possible (to) represent`

tips/TIP-0018/tip-0018.md
```diff
@@ -0,0 +1,3755 @@
...
+</table>
+
+### Additional Transaction Syntactic Validation Rules
+
+- `Amount` field must fulfill the dust protection requirements.
+- `Native Tokens Count` must not be greater than `Max Native Tokens Count`.
+- `Native Tokens` must be lexicographically sorted based on `Token ID`.
+- Each <i>Native Token</i> must be unique in the set of `Native Tokens` based on its `Token ID`. No duplicates are
+  allowed.
+- It must hold true that `0` ≤ `Unlock Conditions Count` ≤ `4`.
```
luca-moser: This condition is incorrect as there must always be an `AddressUnlockCondition`

luca-moser (changes requested):

tips/TIP-0018/tip-0018.md
lzpap: Good catch!

tips/TIP-0018/tip-0018.md
lzpap: indeed

tips/TIP-0018/tip-0018.md
```diff
@@ -0,0 +1,3755 @@
...
+  that deposits `Return Amount` IOTA coins to `Return Address` via an output that has no additional spending
+  constraints. (<i>ExtendedOutput</i> with only an <i>Address Unlock Condition</i>)
+- When several outputs with <i>Dust Deposit Unlock Condition</i> and the same `Return Address` are consumed, their return
+  amounts per `Return Addresses` are summed up and the output side of the transaction must deposit this total sum per
+  `Return Address`.
+
+<details>
+<summary>Dust Deposit Return Unlock Condition</summary>
+<blockquote>
+        Defines the amount of IOTAs used as dust deposit that have to be returned to <i>Sender</i>.
```
thibault-martinez: Not **Sender** anymore

tips/TIP-0018/tip-0018.md
```diff
@@ -0,0 +1,3755 @@
...
+## NFT Output
+
+Non-fungible tokens in the ledger are implemented with a special output type, the so-called <i>NFTOutput</i>.
+
+Each NFT output gets assigned a unique identifier `NFT ID` upon creation by the protocol. `NFT ID` is BLAKE2b-160 hash
+of the <i>Output ID</i>  that created the NFT. The address of the NFT is the concatenation of `NFT Address Type` ||
+`NFT ID`.
+
+The NFT may contain immutable metadata set upon creation, and a verified `Issuer`. The output type supports all
+ unlock conditions and optional feature blocks so that the output can be sent as a request to smart contract chain
```
thibault-martinez: > The output type supports all unlock conditions

Not true, it doesn't support the StateController and Governor conditions

tips/TIP-0018/tip-0018.md
```diff
@@ -0,0 +1,3755 @@
...
+            </details>
+        </td>
+    </tr>
+</table>
+
+#### Additional syntactic output validation rules:
+
+- `Native Tokens` must be lexicographically sorted based on `Token ID`.
+- Each <i>Native Token</i> must be unique in the set of `Native Tokens` based on its `Token ID`. No duplicates are
+  allowed.
```
thibault-martinez: Shouldn't we syntactically check that the amount of any native token in the set is not 0 ?

tips/TIP-0018/tip-0018.md
```diff
@@ -0,0 +1,3755 @@
...
+        <td>uint32</td>
+        <td>
+            Unix time (seconds since Unix epoch) starting from which the output can be consumed.
+        </td>
+    </tr>
+</table>
+
+##### Expiration Unlock Conditions
+
+The expiration feature of outputs makes it possible for the sender to reclaim an output after a given expiration time
```
thibault-martinez: Should be return address instead of sender ?
Same for next paragraph.

tips/TIP-0018/tip-0018.md
```diff
@@ -0,0 +1,3755 @@
...
+
+<table>
+    <tr>
+        <td><b>Name</b></td>
+        <td><b>Type</b></td>
+        <td><b>Description</b></td>
+    </tr>
+    <tr>
+        <td>Native Tokens Count</td>
+        <td>uint16</td>
```
alexsporn: I think uint16 is way too big for the amount of native tokens. uint8 is more than enough. We should even maybe limit it to a smaller number.

tips/TIP-0018/tip-0018.md
```diff
@@ -0,0 +1,3755 @@
...
+| Ed25519 | 0 | iota1**q**... |
+| Alias | 8 | iota1**p**... |
+| NFT | 16 | iota1**z**... |
+
+A user can identify by looking at the address whether it is a signature backed address, a smart contract chain account
+or an NFT address.
+
+##### Dust Deposit Return Unlock Condition
+
+This unlock condition is employed to achieve conditional sending. An output that <i>Dust Deposit Unlock Condition</i>
```
thibault-martinez: > An output that <i>Dust Deposit...

Missing word

tips/TIP-0018/tip-0018.md
```diff
@@ -0,0 +1,3755 @@
...
+        <td>Data</td>
+        <td>ByteArray</td>
+        <td>Binary data.</td>
+    </tr>
+</table>
+
+#### Tag Block
+
+A <i>Tag Block</i> makes it possible to tag outputs with an index, so they can be retrieved through an indexer API not
+only by their address, but also based on the the `Tag`. **The combination of an <i>Tag Block</i>, a
```
thibault-martinez:
```suggestion
only by their address, but also based on the the `Tag`. **The combination of a <i>Tag Block</i>, a
```

oopsmonk (commented):
The definition of the Alias/NFT address fields are incorrect in all Address tables.

tips/TIP-0018/tip-0018.md
```diff
@@ -0,0 +1,3755 @@
...
+- `Amount` field must fulfill the dust protection requirements.
+- `Native Tokens Count` must not be greater than `Max Native Tokens Count`.
+- `Native Tokens` must be lexicographically sorted based on `Token ID`.
+- Each <i>Native Token</i> must be unique in the set of `Native Tokens` based on its `Token ID`. No duplicates are
+  allowed.
+- It must hold true that `Unlock Conditions Count` = `1`.
+- `Unlock Condition Type` of an <i>Unlock Condition</i> must define on of the following types:
+  - <i>Address Unlock Condition</i>
+- <i>Address Unlock Condition</i> might only define an <i>Alias Address</i> as `Address`.
+- It must hold true that `0` ≤ `Blocks Count` ≤ `1`.
```
thibault-martinez: You usually list the possible values but don't do it for this one

tips/TIP-0018/tip-0018.md
```diff
@@ -0,0 +1,3755 @@
...
+corresponding alias) at a smaller index.
+
+For example the scenario where `Alias A` is locked to the address of `Alias B` while `Alias B` is in locked to the
+address of `Alias A` introduces a circular dependency and is not well-defined. By requiring the *Unlock Blocks* to be
+ordered as described above, a transaction consuming `Alias A` as well as `Alias B` can never be valid as there would
+always need to be one *Alias Unlock Block* referencing a greater index.
+
+#### Alias Unlock Block Syntactic Validation
+
+ - It must hold that 0 ≤ `Alias Reference Unlock Index` ≤ `Max Inputs Count`.
```
thibault-martinez: Shouldn't this be `< Max Inputs Count` ?

tips/TIP-0018/tip-0018.md
```diff
@@ -0,0 +1,3755 @@
...
+
+An *Alias Unlock Block* is only valid if the input in the transaction at index `NFT Reference Unlock Index` is the NFT
+output with the same `NFT ID` as the `Address` field of the to-be unlocked output.
+
+If the i-th *Unlock Block* of a transaction is an *NFT Unlock Block* and has `NFT Reference Unlock Index` set to k, it
+must hold that i > k. Hence, an <i>NFT Unlock Block</i> can only reference an *Unlock Block* at a smaller index.
+
+#### NFT Unlock Block Syntactic Validation
+
+- It must hold that 0 ≤ `NFT Reference Unlock Index` ≤ `Max Inputs Count`.
```
thibault-martinez: Same, shouldn't this be `< Max Inputs Count` ?

tips/TIP-0018/tip-0018.md
```diff
@@ -0,0 +1,3755 @@
...
+- <i>Blocks</i> must be sorted in ascending order based on their `Block Type`.
+- Syntactic validation of all present feature blocks must pass.
+
+### Additional Transaction Semantic Validation Rules
+
+#### Consumed Outputs
+
+- The unlock block of the input must correspond to `Address` field in the <i>Address Unlock Condition</i> and the
+  unlock must be valid.
+- The unlock is valid if and only if all unlock conditions and feature blocks present in the output validate.
```
thibault-martinez: Is there actually anything that needs validation regarding feature blocks of consumed outputs ?

tips/TIP-0018/tip-0018.md
```diff
@@ -0,0 +1,3755 @@
...
+    - Governance transition must only change the following fields:
+      `Address` of <i>State Controller Address Unlock Condition</i>,`Address` of <i>Governor Address Unlock Condition</i>,
+      `Metadata Block`, `Sender Block`.
+    - The `Metadata Block` is optional, the governor can put additional info about the chain here, for example chain
+      name, fee structure, supported VMs, list of access nodes, etc., anything that helps clients to fetch info (i.e.
+      account balances) about the layer 2 network.
+- When a consumed alias output has an `Issuer Block` and a corresponding alias output on the output side,
+  `Issuer Block` is not allowed to change.
+
+#### Created Outputs
```
thibault-martinez: The creation constraints of an AliasOutput (zeroed AliasId, state index, foundry counter...) are not very clear/explicit

tips/TIP-0018/tip-0018.md
lzpap: In the next paragraph, sender is used in other context (sender-receiver) not as the name of the field.

tips/TIP-0018/tip-0018.md
lzpap: Yes we should, thanks!

tips/TIP-0018/tip-0018.md
lzpap: you are right!

tips/TIP-0018/tip-0018.md
lzpap: yes

tips/TIP-0018/tip-0018.md
lzpap: Yes, there is one, the issuer block: you need to validate that it will be carried over to the next state of the UTXO state machine.

tips/TIP-0018/tip-0018.md
```diff
@@ -0,0 +1,3762 @@
...
+                    </details>
+                </td>
+            </tr>
+        </table>
+    </details>
+</table>
+
+### Additional Transaction Syntactic Validation Rules
+
+- `Amount` field must fulfill the dust protection requirements.
```
thibault-martinez: For all outputs, maybe this needs a bit of clarification. I think The actual value of the amount is checked against zero in the Transaction TIP but not in this one, which would suggest that an Output with a 0 amount is valid in itself, until it is put in a transaction. I would say it's invalid in the first place.

tips/TIP-0018/tip-0018.md
thibault-martinez: Thanks

tips/TIP-0018/tip-0018.md
lzpap: Indeed, it should be invalid.

tips/TIP-0018/tip-0018.md
lzpap: yes it makes sense. I'll add the `uint8` change and we can discuss in #60 how we should limit it.

tips/TIP-0018/tip-0018.md
```diff
@@ -0,0 +1,3767 @@
...
+- It must hold true that `0` ≤ `Blocks Count` ≤ `3`.
+- `Block Type` of a <i>Block</i> must define on of the following types:
+  - <i>Sender Block</i>
+  - <i>Issuer Block</i>
+  - <i>Metadata Block</i>
+- Syntactic validation of all present feature blocks must pass.
+- When `Alias ID` is zeroed out, `State Index` and `Foundry Counter` must be `0`.
+- `State Metadata Length` must not be greater than `Max Metadata Length`.
+- `Address` of <i>State Controller Address Unlock Condition</i> and `Address` of
+  <i>Governor Address Unlock Condition</i> must be different from the alias address derived from `Alias ID`.
```
thibault-martinez: Maybe my understanding but it's not clearly stated that they are both mandatory

thibault-martinez: Actually, except for Extended, it's not clear there must be an address/state/governor

thibault-martinez: Just saw it's there for NFT sorry 😬

lzpap: > Actually, except for Extended, it's not clear there must be an address/state/governor

Implicitly it is, as it is defined for the alias output that
 -  supported unlock conditions are the state/governor,
 - `Unlock Conditions Count` must be `2`,
 - and there can be at most one of each type of unlock conditions in the output.
 
 This means ther ehas to be one state controller unlock condition and one governor in each alias otuput, otherwise they are syntactically invalid.

thibault-martinez: Yeah right, a bit of mental overhead but that's correct, thanks 👍

tips/TIP-0018/tip-0018.md
```diff
@@ -0,0 +1,3767 @@
...
+##### Dust Deposit Return Unlock Condition
+
+This unlock condition is employed to achieve conditional sending. An output that has <i>Dust Deposit Unlock Condition</i>
+specified can only be consumed in a transaction that deposits `Return Amount` IOTA coins into `Return Address`. When
+several of such outputs are consumed, their return amounts per `Return Addresses` are summed up and the output side
+must deposit this total sum per `Return Address`.
+
+###### Additional syntactic transaction validation rule:
+
+- `Return Amount` must be ≥ than `Minimum Dust Deposit` and must not be `0`.
```
thibault-martinez: `Minimum Dust Deposit` is not in https://github.com/iotaledger/tips/pull/43, is there no minimum anymore or is it missing there ?

tips/TIP-0018/tip-0018.md
```diff
@@ -0,0 +1,3767 @@
...
+
+<table>
+    <tr>
+        <td><b>Name</b></td>
+        <td><b>Type</b></td>
+        <td><b>Description</b></td>
+    </tr>
+    <tr>
+        <td>Native Tokens Count</td>
+        <td>uint8</td>
```
thibault-martinez: I know there is a discussion about limits but as a note: the current maximum number of native tokens of 256 doesn't fit in an u8 anymore it should at least be decreased to 255

tips/TIP-0018/tip-0018.md
```diff
@@ -0,0 +1,3767 @@
...
+
+This unlock condition is employed to achieve conditional sending. An output that has <i>Dust Deposit Unlock Condition</i>
+specified can only be consumed in a transaction that deposits `Return Amount` IOTA coins into `Return Address`. When
+several of such outputs are consumed, their return amounts per `Return Addresses` are summed up and the output side
+must deposit this total sum per `Return Address`.
+
+###### Additional syntactic transaction validation rule:
+
+- `Return Amount` must be ≥ than `Minimum Dust Deposit` and must not be `0`.
+- `Return Amount` in a <i>Dust Deposit Unlock Condition</i> must be ≤ than the required dust deposit of the output.
```
Thoralf-M: What's the reason for this limitation?

lzpap: > `Return Amount` must be ≥ than `Minimum Dust Deposit` and must not be `0`.
Otherwise the recipient can't create the output that returns the dust deposit, as it would violate dust rules. It can not  be 0 because then the unlock condition doesn't have any affect, so why let it be included anyway?

> `Return Amount` in a <i>Dust Deposit Unlock Condition</i> must be ≤ than the required dust deposit of the output.
Ottherwise you would be able to await more iotas than you actually used for covering the dust output, which means you are essentially creating a swap offfer. You can read more about the problems related to this in #42 .

tips/TIP-0018/tip-0018.md
```diff
@@ -0,0 +1,3767 @@
...
+                    </details>
+                </td>
+            </tr>
+        </table>
+    </details>
+</table>
+
+### Additional Transaction Syntactic Validation Rules
+
+- `Amount` field must fulfill the dust protection requirements and must not be `0`.
```
thibault-martinez: To clear out any ambiguity: do we want a maximum for the amount ? I know that as part of the transaction validation, the sum of all outputs must not exceed the total supply, but what about outputs on their own ? Is an output with an amount that exceeds the total supply valid ?

lzpap: Of course such an output makes no sense because the transaction validation will fail immeditely. Therefore, this rule is implicitly there, however not explicitly.

Checking if the amount is greater than max supply is a remnant from the bundle days, where you could have negative `Amount` in transactions, so the bundle could have been balanced even if one transaction had greater `Amount` than max supply.

We could add this extra syntactic check to outputs, but as I wrote above, it is redundant.

tips/TIP-0018/tip-0018.md
Thoralf-M: > Why would you lock more of your funds (Return Amount) into the output than you absolutely have to? What does it achieve?

Because otherwise you have to split your funds to multiple outputs if you don't already have exactly the `Return Amount`+`100i` amount in one output, and if you don't have > 2*deposit amount IOTAs, then you can't do it at all, because the remaining IOTAs wouldn't be enough to cover the byte cost

lzpap: > Because otherwise you have to split your funds to multiple outputs if you don't already have exactly the Return Amount+100i amount in one output

Yes, it is unlikely that you have exactly `Return Amount`+`100i` in one input, so you'll have to split something up anyways. Splitting up but still not locking it into the DDR output is stil ldesried, no?

> and if you don't have > 2*deposit amount IOTAs, then you can't do it at all, because the remaining IOTAs wouldn't be enough to cover the byte cost

Yes, this is true, however unlikely that most users will have such low balance <1Mi. If we want to cover this case, the rule would be `0` < `Amount` - `Return Amount` =< `Minimum Dust Deposit`

tips/TIP-0018/tip-0018.md
```diff
@@ -0,0 +1,3955 @@
...
+                    </details>
+                </td>
+            </tr>
+            <tr>
+                <td>Immutable Blocks Count</td>
+                <td>uint8</td>
+                <td>The number of immutable feature blocks following. Immutable blocks are defined upon deployment of the UTXO state machine and are not allowed to change in any future state transition.</td>
+            </tr>
+            <tr>
+                <td valign="top">Blocks <code>atMostOneOfEach</code></td>
```
Thoralf-M: Duplicate, should be removed here

tips/TIP-0018/tip-0018.md
lzpap: I've added it to the TIP.

tips/TIP-0018/tip-0018.md
```diff
@@ -0,0 +1,3959 @@
...
+
+This unlock condition is employed to achieve conditional sending. An output that has
+<i>Storage Deposit Return Unlock Condition</i> specified can only be consumed in a transaction that deposits
+`Return Amount` IOTA coins into `Return Address`. When several of such outputs are consumed, their return amounts per
+`Return Addresses` are summed up and the output side must deposit this total sum per `Return Address`.
+
+###### Additional syntactic transaction validation rule:
+
+- `Return Amount` must be ≥ than `Minimum Storage Deposit` and must not be `0`.
+- It must hold true, that `0` ≤ `Amount` - `Return Amount` ≤ `Required Dust Deposit of the Output`.
```
grtlr, line 416: If I understand correctly, `Minimum Storage Deposit` and `Required Dust Deposit of the Output` are actually the same? So we could unify this a little bit:

```suggestion
- `Return Amount` must not be `0` and must be ≥ than `Minimum Storage Deposit of the Output`.
- It must hold true, that `0` ≤ `Amount` - `Return Amount` ≤ `Minimum Storage Deposit of the Output`.
```

grtlr: Thanks for clearing this up!

Thoralf-M: > [that's why you also need to inspect the address within the condition to find out the minimum btw](https://github.com/iotaledger/iota.go/blob/3d39187c512978b898364c6b0512bb96b4730a48/output.go#L696).

But there you're missing that the input with the `StorageDepositReturnUnlockCondition` could also have native tokens, which should also be sent back and require a larger deposit then, no?

luca-moser: The idea is to be able to create an output with native tokens where the `StorageDepositReturnUnlockCondition` allows you to specify the iota tokens you get back once the receiver wants to consume the output. Don't think we modelled it to support a use case where you want to have native tokens sent back up on consumption. Can you specify a scenario in which this would come into play?

Thoralf-M: Sorry ignore my comment, I mixed it up with the expiration unlock condition

tips/TIP-0018/tip-0018.md
```diff
@@ -0,0 +1,3973 @@
...
+#### Output Syntactic Validation
+
+- `Amount` field must fulfill the dust protection requirements and must not be `0`.
+- `Amount` field must be ≤ `Max IOTA Supply`.
+- `Native Tokens Count` must not be greater than `Max Native Tokens Count`.
+- `Native Tokens` must be lexicographically sorted based on `Token ID`.
+- Each <i>Native Token</i> must be unique in the set of `Native Tokens` based on its `Token ID`. No duplicates are
+  allowed.
+- `Amount` of any <i>Native Token</i> must not be `0`.
+- It must hold true that `0` ≤ `Unlock Conditions Count` ≤ `4`.
```
oopsmonk, line 3739: > Address Unlock Condition must be present.

therefore unlock conditions count must not be 0
```suggestion
- It must hold true that `1` ≤ `Unlock Conditions Count` ≤ `4`.
```

tips/TIP-0018/tip-0018.md
```diff
@@ -0,0 +1,3973 @@
...
+      - `State Metadata Length`,
+      - `State Metadata`,
+      - `Foundry Counter` and
+      - `Sender Block` in `Blocks`.
+    - `Foundry Counter` field must increase by the number of foundry outputs created in the transaction that map to
+      `Alias ID`. The `Serial Number` fields of the created foundries must be the set of natural numbers that cover the
+       open-ended interval between the previous and next values of the `Foundry Counter` field in the alias output.
+    - The created foundry outputs must be sorted in the list of outputs by their `Serial Number`. Note, that any
+      foundry that maps to `Alias ID` and has a `Serial Number` that is less or equal to the `Foundry Counter` of the
+      input alias is ignored when it comes to sorting.
```
thibault-martinez, line 2620: What does it mean ignored ? They can be literally anywhere in the outputs ? Even in the middle of the sorted created foundries ?

tips/TIP-0018/tip-0018.md
```diff
@@ -0,0 +1,3973 @@
...
+      - `Native Tokens`,
+      - `State Index`,
+      - `State Metadata Length`,
+      - `State Metadata`,
+      - `Foundry Counter` and
+      - `Sender Block` in `Blocks`.
+    - `Foundry Counter` field must increase by the number of foundry outputs created in the transaction that map to
+      `Alias ID`. The `Serial Number` fields of the created foundries must be the set of natural numbers that cover the
+       open-ended interval between the previous and next values of the `Foundry Counter` field in the alias output.
+    - The created foundry outputs must be sorted in the list of outputs by their `Serial Number`. Note, that any
```
thibault-martinez, line 2618: More generally, does `sorted` here means that nothing can be between them ?

tips/TIP-0018/tip-0018.md
lzpap: Yes, they can be anywhere. The only important thing is foundries **created** in the transaction are sorted. Otherwise we would not be able to check the monotonicity and uniqueness of the serial numbers.

tips/TIP-0018/tip-0018.md
lzpap: No, `sorted` here means that they have to be sorted wrt each other. Then we can check monotonicity and uniqueness easy while iterating over the list of outputs from the beginning.

thibault-martinez: So just to be sure, something like this would be fine? : Created Foundry (6), Basic, Transitioned Foundry (4), Alias, Created Foundry (7), Created Foundry (8), Basic

lzpap: Yes, this would be fine, assuming InputAlias(FoundryCounter) = 5 and outputAlias(FoundryCounter) = 8

thibault-martinez: Thanks 👍

tips/TIP-0018/tip-0018.md
```diff
@@ -0,0 +1,3973 @@
...
+  - <i>Immutable Alias Address Unlock Condition</i>
+- It must hold true that `0` ≤ `Blocks Count` ≤ `1`.
+- `Block Type` of a <i>Block</i> in `Blocks` must define on of the following types:
+  - <i>Metadata Block</i>
+- It must hold true that `0` ≤ `Immutable Blocks Count` ≤ `1`.
+- `Block Type` of a <i>Block</i> in `Immutable Blocks` must define on of the following types:
+  - <i>Metadata Block</i>
+- Syntactic validation of all present feature blocks must pass.
+- `Token Scheme Type` must match one of the supported schemes. Any other value results in invalid output.
+- `Minted Tokens` must not be greater than `Maximum Supply`.
```
luca-moser, line 2938: This particular rule limits the foundry to not be able to further mint tokens once this counter hits the `MaximumSupply`. However, `MaximumSupply` implies the maximum supply that can theoretically be in circulation, meaning that that foundry should always be able to mint new tokens as long as `Minted`-`Melted` < `MaximumSupply`. This is unlike how the foundry worked previously where the `CirculatingSupply` just had to be within bounds of the `MaximumSupply` for the foundry to further mint new tokens.

Note that lifting this rule makes it so that `Minted` and `Melted` token counters can be > `MaximumSupply` but that should be fine as these counters are just indicative for the actual circulating supply. Lifting this rule also makes these counters unbound but they can never be bigger than uint256 anyway (though this is a syntactical check).

If that isn't wanted, then I suggest we rename `MaximumSupply` to `MaxMintedTokens` as to reflect this limit more clearly, however I don't know the rational of why the foundry would need to be limited in such a way.

tips/TIP-0018/tip-0018.md
```diff
@@ -0,0 +1,3973 @@
...
+- It must hold true, that `0` ≤ `Amount` - `Return Amount` ≤ `Required Storage Deposit of the Output`.
+
+###### Additional semantic transaction validation rule:
+
+- An output that has <i>Storage Deposit Return Unlock Condition</i> specified must only be consumed and unlocked in a
+  transaction that deposits `Return Amount` IOTA coins to `Return Address` via an output that has no additional spending
+  constraints. (<i>Basic Output</i> with only an <i>Address Unlock Condition</i>)
+- When several outputs with <i>Storage Deposit Return Unlock Condition</i> and the same `Return Address` are consumed,
+  their return amounts per `Return Addresses` are summed up and the output side of the transaction must deposit this
+  total sum per `Return Address`.
```
thibault-martinez, line 426: It is not fully clear, to me, if the summed deposit should be done in a single BasicOutput or not

tips/TIP-0018/tip-0018.md
```diff
@@ -0,0 +1,3973 @@
...
+###### Additional syntactic transaction validation rule:
+
+- `Return Amount` must be ≥ than `Minimum Storage Deposit` and must not be `0`.
+- It must hold true, that `0` ≤ `Amount` - `Return Amount` ≤ `Required Storage Deposit of the Output`.
+
+###### Additional semantic transaction validation rule:
+
+- An output that has <i>Storage Deposit Return Unlock Condition</i> specified must only be consumed and unlocked in a
+  transaction that deposits `Return Amount` IOTA coins to `Return Address` via an output that has no additional spending
+  constraints. (<i>Basic Output</i> with only an <i>Address Unlock Condition</i>)
```
thibault-martinez, line 423: Just to be sure, is it allowed to deposit native tokens?

lzpap: Yes, it is, it just shouldn't have any other unlock conditions than `Address Unlock` on the basic output(s).

tips/TIP-0018/tip-0018.md
lzpap: No, it can be done in several as well. Will add clarification here :+1:

tips/TIP-0018/tip-0018.md
lzpap: The first option, `Minted`-`Melted` < `MaximumSupply`, is now added to Token Scheme Validation in   https://github.com/iotaledger/tips/pull/38/commits/4257b772cab90178832ad7f43fa60d6cdd1ddc85

tips/TIP-0018/tip-0018.md
```diff
@@ -0,0 +1,3900 @@
...
+| NFT | 16 | iota1**z**... |
+
+A user can identify by looking at the address whether it is a signature backed address, a smart contract chain account
+or an NFT address.
+
+##### Storage Deposit Return Unlock Condition
+
+This unlock condition is employed to achieve conditional sending. An output that has
+<i>Storage Deposit Return Unlock Condition</i> specified can only be consumed in a transaction that deposits
+`Return Amount` IOTA coins into `Return Address`. When several of such outputs are consumed, their return amounts per
```
thibault-martinez, line 411: The second sentence is kinda redundant with the one in the semantic rules.

tips/TIP-0018/tip-0018.md
```diff
@@ -0,0 +1,3900 @@
...
+###### Additional semantic transaction validation rule:
+
+- An output that has <i>Storage Deposit Return Unlock Condition</i> specified must only be consumed and unlocked in a
+  transaction that deposits `Return Amount` IOTA coins to `Return Address` via one or more outputs that:
+   - are of type [<i>Basic Output</i>](#basic-output),
+   - have only an [<i>Address Unlock Condition</i>](#address-unlock-condition) defined, and
+   - have no [<i>Feature Blocks</i>](#feature-blocks).
+- When several outputs with <i>Storage Deposit Return Unlock Condition</i> and the same `Return Address` are consumed,
+  their return amounts per `Return Addresses` are summed up and the output side of the transaction must deposit this
+  total sum per `Return Address` via output(s) that satisfy the previous condition.
```
thibault-martinez, line 428: Should the total sum in deposit be absolutely equal to the storage returns sum or can it be greater or equal ?

lzpap: It can be greater or equal with https://github.com/iotaledger/tips/pull/38/commits/89b0c1b16f253d77e860472878d93f3dd7678764

tips/TIP-0018/tip-0018.md
```diff
@@ -0,0 +1,3902 @@
...
+##### Simple Token Scheme Validation Rules
+- Let `Token Diff` denote the **difference between native token balances of the input and the output side** of the
+  transaction of the single `Token ID` that maps to the `Foundry ID`. Minting results in excess of tokens on the output
+  side (positive diff), melting results in excess on the input side (negative diff). Now, the following conditions must
+  hold for `Token Diff`:
+  1. When `Token Diff` > 0
+    - `Current State(Minted Tokens) + Token Diff = Next State(Minted Tokens)`.
+    - `Current State(Melted Tokens) = Next State(Melted Tokens)`
+  2. When `Token Diff` < 0, it must hold true that:
+    - `Current State(Melted Tokens) <= Next State(Melted Tokens)`
```
thibault-martinez, line 2941: Can they actually be equal if `Token Diff < 0` ?

lzpap:
> ISCP and dRNG (on Goshimmer) uses thresold singature. I'd suggest to add an address type derived from a BLS public key to allow committees of nodes that generate a DKG based public key to sing L1 txs.  This signature are not currently considered valid by the network. 

ISC will be using Threshold Eddsa signatures on Stardust and onwards, because they are more efficient, properly vetted and validation is already supported by the L1 network since chrysalis.

Committees will be generating threshold Eddsa public keys via DKG and are capable of signing L1 transactions.

There is no need to add a new and heavier signature scheme to L1 if we can already fulfill the functional requirements of threshold signatures.

sdellava:
Perfect. Thanks.

tips/TIP-0018/tip-0018.md
```diff
@@ -0,0 +1,3832 @@
...
+###### Additional semantic transaction validation rule:
+
+- An output that has <i>Storage Deposit Return Unlock Condition</i> specified must only be consumed and unlocked in a
+  transaction that deposits `Return Amount` IOTA coins to `Return Address` via one or more outputs that:
+   - are of type [<i>Basic Output</i>](#basic-output),
+   - have only an [<i>Address Unlock Condition</i>](#address-unlock-condition) defined, and
+   - have no [<i>Features</i>](#features).
+- When several outputs with <i>Storage Deposit Return Unlock Condition</i> and the same `Return Address` are consumed,
+  their return amounts per `Return Addresses` are summed up and the output side of the transaction must deposit
+  _at least_ this total sum per `Return Address` via output(s) that satisfy the previous condition.
```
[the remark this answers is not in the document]
nicole-obrien: > We don't allow additional unlock conditions

Ah yes I meant to write this, I agree that you shouldn't have allow unlock conditions.

> Features don't affect the unlocking so we could theoretically allow them.

This is what I think we shouldn't limit

> If you used the SDRUC however, it means you want to send something which is less than the storage deposit, or its a native token. So if the recipient wants to return it with features, they have to pay into the output since features increase the storage deposit.

If the original output (with the SDRUC) had features then it would have contained enough IOTA to cover the storage deposit, hence you can inherently send it back without needing to add any of your own.

If the original output didn't have features then yes you may need to provide more IOTA to cover the storage deposit, but thats up to the sender to decide, I don't think we should limit too much what could be a valid UTXO.

> If you really wanted we could also allow NFTs with only an Address Unlock Condition, but does it make sense?

Yes I think it makes sense to be able to return an NFT it your don't want it. And again its the freedom of the option to do this, as I don't see a technical limitation of allowing it.

> So even if you don't have funds in your wallet, you can satisfy the SDRUC unlock conditions if you send back a basic output with X+Y tokens to return address. (btw in practice I think the receiver would just pocket Y tokens for themselves)

The receiver couldn't pocket the tokens if they couldn't provide the amount to satisfy the unlock.

tips/TIP-0018/tip-0018.md
```diff
@@ -0,0 +1,3832 @@
...
+    - **For created outputs**: validates whether the output can be created in a transaction given the semantic
+      transaction context.
+
+Each new output type may add its own validation rules which become part of the transaction validation rules if the
+output is placed inside a transaction. <i>Unlock Conditions</i> and <i>Features</i> described previously also add
+constraints to transaction validation when they are placed in outputs.
+
+## Basic Output
+
+<i>Basic Output</i> can hold native tokens and might have several unlock conditions and optional featuress. The
```
luca-moser, line 1341:
```suggestion
<i>Basic Output</i> can hold native tokens and might have several unlock conditions and optional features. The
```

tips/TIP-0018/tip-0018.md
```diff
@@ -0,0 +1,3832 @@
...
+- `Unlock Condition Type` of an <i>Unlock Condition</i> must define on of the following types:
+  - <i>Address Unlock Condition</i>
+  - <i>Storage Deposit Return Unlock Condition</i>
+  - <i>Timelock Unlock Condition</i>
+  - <i>Expiration Unlock Condition</i>
+- <i>Unlock Conditions</i> must be sorted in ascending order based on their `Unlock Condition Type`.
+- Syntactic validation of all present unlock conditions must pass.
+- <i>Address Unlock Condition</i> must be present.
+- It must hold true that `0` ≤ `Features Count` ≤ `3`.
+- `Feature Type` of a <i>Feature</i> must define on of the following types:
```
luca-moser, line 1912:
```suggestion
- `Feature Type` of a <i>Feature</i> must define one of the following types:
```

tips/TIP-0018/tip-0018.md
```diff
@@ -0,0 +1,3832 @@
...
+  allowed.
+- `Amount` of any <i>Native Token</i> must not be `0`.
+- It must hold true that `Unlock Conditions Count` = `2`.
+- `Unlock Condition Type` of an <i>Unlock Condition</i> must define one of the following types:
+  - <i>State Controller Address Unlock Condition</i>
+  - <i>Governor Address Unlock Condition</i>
+- <i>Unlock Conditions</i> must be sorted in ascending order based on their `Unlock Condition Type`.
+- Syntactic validation of all present unlock conditions must pass.
+- It must hold true that `0` ≤ `Features Count` ≤ `2`.
+- `Feature Type` of a <i>Feature</i> in `Features` must define on of the following types:
```
luca-moser, line 2481:
```suggestion
- `Feature Type` of a <i>Feature</i> in `Features` must define one of the following types:
```

tips/TIP-0018/tip-0018.md
```diff
@@ -0,0 +1,3832 @@
...
+- `Native Tokens` must be lexicographically sorted based on `Token ID`.
+- Each <i>Native Token</i> must be unique in the set of `Native Tokens` based on its `Token ID`. No duplicates are
+  allowed.
+- `Amount` of any <i>Native Token</i> must not be `0`.
+- It must hold true that `Unlock Conditions Count` = `1`.
+- `Unlock Condition Type` of an <i>Unlock Condition</i> must define on of the following types:
+  - <i>Immutable Alias Address Unlock Condition</i>
+- Syntactic validation of all present unlock conditions must pass.
+- It must hold true that `0` ≤ `Features Count` ≤ `1`.
+- `Feature Type` of a <i>Feature</i> in `Features` must define on of the following types:
```
luca-moser, line 2820:
```suggestion
- `Feature Type` of a <i>Feature</i> in `Features` must define one of the following types:
```

tips/TIP-0018/tip-0018.md
```diff
@@ -0,0 +1,3832 @@
...
+
+- `Amount` field must fulfill the dust protection requirements and must not be `0`.
+- `Amount` field must be ≤ `Max IOTA Supply`.
+- `Native Tokens Count` must not be greater than `Max Native Tokens Count`.
+- `Native Tokens` must be lexicographically sorted based on `Token ID`.
+- Each <i>Native Token</i> must be unique in the set of `Native Tokens` based on its `Token ID`. No duplicates are
+  allowed.
+- `Amount` of any <i>Native Token</i> must not be `0`.
+- It must hold true that `Unlock Conditions Count` = `1`.
+- `Unlock Condition Type` of an <i>Unlock Condition</i> must define on of the following types:
```
luca-moser, line 2816:
```suggestion
- `Unlock Condition Type` of an <i>Unlock Condition</i> must define one of the following types:
```

tips/TIP-0018/tip-0018.md
```diff
@@ -0,0 +1,3832 @@
...
+  - <i>Immutable Alias Address Unlock Condition</i>
+- Syntactic validation of all present unlock conditions must pass.
+- It must hold true that `0` ≤ `Features Count` ≤ `1`.
+- `Feature Type` of a <i>Feature</i> in `Features` must define on of the following types:
+  - <i>Metadata Feature</i>
+- It must hold true that `0` ≤ `Immutable Features Count` ≤ `1`.
+- `Feature Type` of a <i>Feature</i> in `Immutable Features` must define on of the following types:
```
luca-moser, line 2823:
```suggestion
- `Feature Type` of a <i>Feature</i> in `Immutable Features` must define one of the following types:
```

tips/TIP-0018/tip-0018.md
```diff
@@ -0,0 +1,3832 @@
...
+
+- `Amount` field must fulfill the dust protection requirements and must not be `0`.
+- `Amount` field must be ≤ `Max IOTA Supply`.
+- `Native Tokens Count` must not be greater than `Max Native Tokens Count`.
+- `Native Tokens` must be lexicographically sorted based on `Token ID`.
+- Each <i>Native Token</i> must be unique in the set of `Native Tokens` based on its `Token ID`. No duplicates are
+  allowed.
+- `Amount` of any <i>Native Token</i> must not be `0`.
+- It must hold true that `1` ≤ `Unlock Conditions Count` ≤ `4`.
+- `Unlock Condition Type` of an <i>Unlock Condition</i> must define on of the following types:
```
luca-moser, line 1903:
```suggestion
- `Unlock Condition Type` of an <i>Unlock Condition</i> must define one of the following types:
```

tips/TIP-0018/tip-0018.md
```diff
@@ -0,0 +1,3832 @@
...
+  While it is not necessarily needed for the protocol, it is needed for client side operations, such as:
+    - Retrieving the current state of the foundry.
+    - Accessing token metadata in foundry based on `Foundry ID`/`Tokend ID`.
+
+## NFT Output
+
+Non-fungible tokens in the ledger are implemented with a special output type, the so-called <i>NFTOutput</i>.
+
+Each NFT output gets assigned a unique identifier `NFT ID` upon creation by the protocol. `NFT ID` is BLAKE2b-256 hash
+of the <i>Output ID</i>  that created the NFT. The address of the NFT is the concatenation of `NFT Address Type` ||
```
luca-moser, line 2915:
```suggestion
of the <i>Output ID</i> that created the NFT. The address of the NFT is the concatenation of `NFT Address Type` ||
```

tips/TIP-0018/tip-0018.md
```diff
@@ -0,0 +1,3832 @@
...
+
+- `Amount` field must fulfill the dust protection requirements and must not be `0`.
+- `Amount` field must be ≤ `Max IOTA Supply`.
+- `Native Tokens Count` must not be greater than `Max Native Tokens Count`.
+- `Native Tokens` must be lexicographically sorted based on `Token ID`.
+- Each <i>Native Token</i> must be unique in the set of `Native Tokens` based on its `Token ID`. No duplicates are
+  allowed.
+- `Amount` of any <i>Native Token</i> must not be `0`.
+- It must hold true that `1` ≤ `Unlock Conditions Count` ≤ `4`.
+- `Unlock Condition Type` of an <i>Unlock Condition</i> must define on of the following types:
```
luca-moser, line 3599:
```suggestion
- `Unlock Condition Type` of an <i>Unlock Condition</i> must define one of the following types:
```

tips/TIP-0018/tip-0018.md
```diff
@@ -0,0 +1,3832 @@
...
+  - <i>Address Unlock Condition</i>
+  - <i>Storage Deposit Return Unlock Condition</i>
+  - <i>Timelock Unlock Condition</i>
+  - <i>Expiration Unlock Condition</i>
+- <i>Unlock Conditions</i> must be sorted in ascending order based on their `Unlock Condition Type`.
+- Syntactic validation of all present unlock conditions must pass.
+- <i>Address Unlock Condition</i> must be present.
+- It must hold true that `0` ≤ `Features Count` ≤ `3`.
+- `Feature Type` of a <i>Feature</i> in `Features` must define on of the following types:
```
luca-moser, line 3608:
```suggestion
- `Feature Type` of a <i>Feature</i> in `Features` must define one of the following types:
```

tips/TIP-0018/tip-0018.md
```diff
@@ -0,0 +1,3832 @@
...
+  - <i>Sender Feature</i>
+  - <i>Metadata Feature</i>
+  - <i>Tag Feature</i>
+- It must hold true that `0` ≤ `Immutable Features Count` ≤ `2`.
+- `Feature Type` of a <i>Feature</i> in `Immutable Features` must define on of the following types:
```
luca-moser, line 3613:
```suggestion
- `Feature Type` of a <i>Feature</i> in `Immutable Features` must define one of the following types:
```

luca-moser (changes requested):

tips/TIP-0018/tip-0018.md
```diff
@@ -0,0 +1,3832 @@
...
+
+This unlock condition is employed to achieve conditional sending. An output that has
+<i>Storage Deposit Return Unlock Condition</i> specified can only be consumed in a transaction that deposits
+`Return Amount` IOTA coins into `Return Address`. When several of such outputs are consumed, their return amounts per
+`Return Addresses` are summed up and the output side must deposit this total sum per `Return Address`.
+
+###### Additional syntactic transaction validation rule:
+
+- `Minimum Storage Deposit` is the storage deposit in the base currency required for a <i>Basic Output</i> that only
+  has an <i>Address Unlock Condition</i> with `Return Address`, no additional unlock conditions, no features and
```
Thoralf-M, line 417:
```suggestion
  has an <i>Address Unlock Condition</i>, no additional unlock conditions, no features and
```

lzpap: `Return Address` here refers to the field of SDRUC of the consumed output, so I don't get why this should be removed.

Thoralf-M: To me it sounds just like a definition for `Minimum Storage Deposit`, which is unrelated to the SDRUC

lzpap: Alright, if you say so :cat:

luca-moser (approved):

thibault-martinez (approved):

Thoralf-M (approved):

PR #50, opened 2022-01-17
Update TIP-6: Tagged data payload

Wollac:
This PR removes the _Indexation Payload_ which was used to index data in the nodes by a _Tagged Data Payload_ that is not indexed.
This PR modifies the existing TIP-6.


lzpap:
With #55 , this PR should be closed, right?

PR #54, opened 2022-01-24
TIP-23 Tagged Data Payload

Wollac:
This PR creates a new TIP containing the Tagged Data Payload.
It supersedes #50 

[Rendered Version](https://github.com/Wollac/protocol-rfcs/blob/tagged-data/tips/TIP-0023/tip-0023.md)

### DIFF
--- a/tips/TIP-0023/tip-0023.md
+++ b/tips/TIP-0023/tip-0023.md
@@ -24,18 +24,17 @@ The most flexible way to extend an existing object is by the addition of arbitra
 
 The following table describes the serialization of a _Tagged Data Payload_ following the notation from [TIP-21](../TIP-0021/tip-0021.md):
 
-| Name         | Type                  | Description                                          |
-| ------------ | --------------------- | ---------------------------------------------------- |
-| Payload Type | uint32                | Set to *value 5* to denote an _Tagged Data Payload_. |
-| Tag Length   | uint8                 | The length of the following tag field in bytes.      |
-| Tag          | ByteArray[Tag Length] | The tag of the data.                                 |
-| Data         | ByteArray             | Binary data.                                         |
+| Name         | Type              | Description                                              |
+|--------------|-------------------|----------------------------------------------------------|
+| Payload Type | uint32            | Set to *value 5* to denote an _Tagged Data Payload_.     |
+| Tag          | (uint8)ByteArray  | The tag of the data. A leading uint8 denotes its length. |
+| Data         | (uint32)ByteArray | Binary data. A leading uint32 denotes its length.        |
 
 It is important to note that `Tag` is not considered by the protocol, it just serves as a marker for second layer applications.
 
 ## Syntactic Validation
 
-- `Tag Length` must not be larger than `Max Tag Length`. When `Tag Length` is 0, the `Tag` has length 0 and is omitted.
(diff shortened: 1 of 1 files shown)


tips/TIP-0023/tip-0023.md
```diff
@@ -0,0 +1,47 @@
...
+| Payload Type | uint32                | Set to *value 5* to denote an _Tagged Data Payload_. |
+| Tag Length   | uint8                 | The length of the following tag field in bytes.      |
+| Tag          | ByteArray[Tag Length] | The tag of the data.                                 |
+| Data         | ByteArray             | Binary data.                                         |
+
+It is important to note that `Tag` is not considered by the protocol, it just serves as a marker for second layer applications.
+
+## Syntactic Validation
+
+- `Tag Length` must not be larger than `Max Tag Length`. When `Tag Length` is 0, the `Tag` has length 0 and is omitted.
```
lzpap, line 38: Let's reference `Max Tag Length` from TIP-22

tips/TIP-0023/tip-0023.md
```diff
@@ -0,0 +1,47 @@
...
+## Serialized Layout
+
+The following table describes the serialization of a _Tagged Data Payload_ following the notation from [TIP-21](../TIP-0021/tip-0021.md):
+
+| Name         | Type                  | Description                                          |
+| ------------ | --------------------- | ---------------------------------------------------- |
+| Payload Type | uint32                | Set to *value 5* to denote an _Tagged Data Payload_. |
+| Tag Length   | uint8                 | The length of the following tag field in bytes.      |
+| Tag          | ByteArray[Tag Length] | The tag of the data.                                 |
+| Data         | ByteArray             | Binary data.                                         |
```
oopsmonk, line 32: how do we know the length of the arbitrary data, shouldn't have a `Data Length` field? 
Is max data length depend on parent number and tag length?

oopsmonk: according to TIP-18, all ByteArray paired with a length field in structures. for instance `State Metadata` has `State Metadata Length` in the [Alias output](https://github.com/lzpap/tips/blob/master/tips/TIP-0018/tip-0018.md#alias-output). serialization of State Metadata would be `uint16 + uint32 + data bytes` which is not correct. 
the definition of ByteArray in TIP-21 is not true since the length limitation of ByteArray is vary depends on the object.

alexsporn:
@lzpap we currently allow payloads with empty tag and empty data. Is this correct?

lzpap:
> @lzpap we currently allow payloads with empty tag and empty data. Is this correct?

For the indexation payload we allowed empty data, but not empty tag becasue you needed something to index with. Now that we don't do indexing with the Tagged Data Payload, we don't need to enforce the tag IMO.

I see value in having empty payload, for example if you just want to issue messages to help confirmation, but you want to do the absolute minimum PoW for the msg.

What do you think @alexsporn?

Thoralf-M:
@lzpap for absolute minimum PoW one would just use no payload

lzpap:
> @lzpap for absolute minimum PoW one would just use no payload

yes, you are right. The `oneOf` keyword mislead me in the msg structure, it should be rather `optOneOf`.
image

lzpap (approved):

alexsporn (approved):

luca-moser (approved):

PR #57, opened 2022-01-26
TIP-25 Node Core REST API

lzpap:
Core API specification update for Stardust, mostly based on [TIP-13](https://github.com/iotaledger/tips/blob/main/tips/TIP-0013/tip-0013.md).

Important updates:
 - Renaming _messages_ to _blocks_
 - Removal of block lookup by indexation tag. Should be done via a block indexer node plugin that exposes its own routes.
- Removal of address/balance lookups under the utxo routes. Outputs can only be fetched by `OutputID` from the core API. An indexer API described in #53 will expose new routes to fetch outputs based on parameters. 
- Added outputs and unlocks from [TIP-18](#38).
- Added `milestoneIndexBooked` and `milestoneTimestampBooked` to `OutputResponse`.

[Rendered version](https://editor.swagger.io/?url=https://raw.githubusercontent.com/iotaledger/tips/stardust-api/tips/TIP-0025/core-rest-api.yaml)

### DIFF
--- a/tips/TIP-0025/core-rest-api.yaml
+++ b/tips/TIP-0025/core-rest-api.yaml
@@ -1665,6 +1665,7 @@ components:
         type: 7
         index: 15465
         timestamp: 1602227215
+        protocolVersion: 2
         previousMilestoneId: "0x7ad3d67fc7b619e72e588f51fef2379e43e6e9a856635843b3f29aa3a3f1f006"
         parents:
          - "0x7ed3d67fc7b619e72e588f51fef2379e43e6e9a856635843b3f29aa3a3f1f006"
@@ -2480,6 +2481,9 @@ components:
         timestamp:
           type: integer
           description: The Unix timestamp at which the milestone was issued. The unix timestamp is specified in seconds.
+        protocolVersion:
+          type: integer
+          description: Protocol version of the Milestone Payload and its encapsulating block.
         previousMilestoneId:
           type: string
           description: The Milestone ID of the milestone with Index Number - 1.
@@ -2513,6 +2517,7 @@ components:
         - type
         - index
         - timestamp
+        - protocolVersion
         - previousMilestoneId
         - parents
         - inclusionMerkleRoot
--- a/tips/TIP-0025/core-rest-api.yaml
+++ b/tips/TIP-0025/core-rest-api.yaml
@@ -1992,8 +1992,6 @@ components:
       required:
         - type
         - amount
-        - nativeTokens
-        - featureBlocks
         - unlockConditions
 
     AliasOutput:
@@ -2049,12 +2047,9 @@ components:
       required:
         - type
         - amount
-        - nativeTokens
         - aliasId
         - stateIndex
         - foundryCounter
-        - featureBlocks
-        - immutableFeatureBlocks
         - unlockConditions
 
     FoundryOutput:
@@ -2106,13 +2101,10 @@ components:
       required:
         - type
         - amount
-        - nativeTokens
         - serialNumber
         - tokenTag
         - tokenScheme
         - unlockConditions
-        - featureBlocks
-        - immutableFeatureBlocks
 
     NFTOutput:
       description: escribes an NFT output, a globally unique token with metadata attached.
@@ -2161,11 +2153,8 @@ components:
       required:
         - type
         - amount
-        - nativeTokens
         - nftId
         - unlockConditions
-        - featureBlocks
-        - immutableFeatureBlocks
 
     NativeToken:
       description: A native token and its balance in the output.
--- a/tips/TIP-0025/core-rest-api.yaml
+++ b/tips/TIP-0025/core-rest-api.yaml
@@ -1725,6 +1725,9 @@ components:
         amount:
           type: string
           description: Amount of native tokens (up to uint256). Hex-encoded number with 0x prefix.
+      required:
+        - tokenId
+        - amount
 
     Ed25519Address:
       description: The Ed25519 address.
@@ -1776,6 +1779,9 @@ components:
             - $ref: '#/components/schemas/Ed25519Address'
             - $ref: '#/components/schemas/AliasAddress'
             - $ref: '#/components/schemas/NFTAddress'
+      required:
+        - type
+        - address
 
     ImmutableAliasAddressUnlockCondition:
       description: Can be unlocked by unlocking the permanent alias address.
@@ -1787,6 +1793,9 @@ components:
         address:
           oneOf:
             - $ref: '#/components/schemas/AliasAddress'
+      required:
+        - type
+        - address
 
     StorageDepositReturnUnlockCondition:
       description: Can be unlocked by depositing return amount to return address via an output that only has Address Unlock Condition.
@@ -1802,6 +1811,10 @@ components:
         returnAmount:
           type: string
           description: Amount of IOTA tokens the consuming transaction should deposit to the address defined in Return Address. Plain string encoded number.
+      required:
+        - type
+        - returnAddress
+        - returnAmount
 
     TimelockUnlockCondition:
       description: Can be unlocked if the confirming milestone has a >= Milestone Index/Unixt Timestamp.
@@ -1815,6 +1828,8 @@ components:
         unixTime:
           type: integer
           description: Unix time (seconds since Unix epoch) starting from which the output can be consumed.
+      required:
+        - type
 
     ExpirationUnlockCondition:
       description: Defines a milestone index and/or unix time until which only Address, defined in Address Unlock Condition, is allowed to unlock the output. After the milestone index and/or unix time, only Return Address can unlock it.
@@ -1833,6 +1848,9 @@ components:
         unixTime:
           type: integer
           description: Before this unix time, Address Unlock Condition is allowed to unlock the output, after that only the address defined in Return Address.
+      required:
+        - type
+        - returnAddress
 
     StateControllerAddressUnlockCondition:
       description: Can be unlocked by unlocking the address.
@@ -1845,6 +1863,9 @@ components:
             - $ref: '#/components/schemas/Ed25519Address'
             - $ref: '#/components/schemas/AliasAddress'
             - $ref: '#/components/schemas/NFTAddress'
+      required:
+        - type
+        - address
 
     GovernorAddressUnlockCondition:
       description: Can be unlocked by unlocking the address.
@@ -1857,6 +1878,9 @@ components:
             - $ref: '#/components/schemas/Ed25519Address'
             - $ref: '#/components/schemas/AliasAddress'
             - $ref: '#/components/schemas/NFTAddress'
+      required:
+        - type
+        - address
 
     SenderBlock:
       description: Identifies the validated sender of the output.
@@ -1869,6 +1893,9 @@ components:
             - $ref: '#/components/schemas/Ed25519Address'
             - $ref: '#/components/schemas/AliasAddress'
             - $ref: '#/components/schemas/NFTAddress'
+      required:
+        - type
+        - sender
 
     IssuerBlock:
       description: Identifies the validated issuer of the UTXO state machine (alias/NFT).
@@ -1876,11 +1903,14 @@ components:
         type:
           type: integer
           description: Set to value 1 to denote an Issuer Block.
-        sender:
+        issuer:
           oneOf:
             - $ref: '#/components/schemas/Ed25519Address'
             - $ref: '#/components/schemas/AliasAddress'
             - $ref: '#/components/schemas/NFTAddress'
+      required:
+        - type
+        - issuer
 
     MetadataBlock:
       description: Defines metadata (arbitrary binary data) that will be stored in the output.
@@ -1891,6 +1921,9 @@ components:
         data:
           type: string
           description: Hex-encoded binary data with 0x prefix.
+      required:
+        - type
+        - data
 
     TagBlock:
       description: Defines an indexation tag to which the output can be indexed by additional node plugins.
@@ -1901,6 +1934,9 @@ components:
         tag:
           type: string
           description: Hex-encoded binary indexation tag with 0x prefix.
+      required:
+        - type
+        - tag
 
     SimpleTokenScheme:
       description: Defines the simple supply control scheme of native tokens. Tokens can be minted by the foundry without additional restrictions as long as maximum supply is requested and circulating supply is not negative.
@@ -2015,13 +2051,18 @@ components:
           description: 256-bit hash based on the message IDs of all the not-ignored state-mutating transactions referenced by the milestone. Hex-encoded with 0x prefix.
         nextPoWScore:
           type: number
+          description: The next minimum PoW score to use after NextPoWScoreMilestoneIndex is hit.
         nextPoWScoreMilestoneIndex:
+          description: The milestone index at which the PoW score changes to NextPoWScore.
           type: number
-        publicKeys:
-          type: array
-          items:
-            type: string
-          description: An array of public keys to validate the signatures. The keys must be in lexicographical order. Hex-encoded with 0x prefix.
+        metadata:
+          type: string
+          description: Hex-encoded binary data with 0x prefix.
+        receipt:
+          type: object
+          description: The inner payload of the milestone. Can be nil or a Receipt.
+          oneOf:
+            - $ref: '#/components/schemas/ReceiptPayload'
         signatures:
           type: array
           items:
@@ -2221,6 +2262,9 @@ components:
     ReceiptPayload:
       description: Contains a receipt and the index of the milestone which contained the receipt.
       properties:
+        type:
+          type: integer
+          description: Type identifier of a receipt payload (3).
         migratedAt:
           type: integer
         final:
@@ -2236,6 +2280,7 @@ components:
         - final
         - funds
         - transaction
+        - type
 
     MigratedFundsEntry:
       properties:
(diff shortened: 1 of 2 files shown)


rufsam:
With https://github.com/iotaledger/iota.go/issues/313, `networkId` from `TransactionEssence` is now of type `string`.
The schemas and examples need to be updated.

tips/TIP-0025/core-rest-api.yaml
```diff
@@ -0,0 +1,2647 @@
...
+        parentMessageIds:
+          - 6c5bc96948813faaf6d5a5f0c6bc453bf2e5343664884aea2873041bdc47c4ca
+          - 97e2c4fce0f6e1023c19ed634e37a44744e21a743a79e8f1fc16d31dc8f47481
+          - c107c5f3d6189954c9b1b8e2fa556db4bcf7ad2da717d328ecf6d9a841a5f37d
+          - ea19397d2670f904c4d8c8616ef7a0724321512f1c09f8eb2aeb0cc4299e2441
+        payload:
+          type: 0
+          essence:
+            type: 0
+            networkId: 9466822412763346725
```
rufsam: With https://github.com/iotaledger/iota.go/issues/313, `networkId` from `TransactionEssence` is now of type `string`.
The schemas and examples need to be updated.

rufsam (changes requested):

tips/TIP-0025/core-rest-api.yaml
lzpap: Done

tips/TIP-0025/core-rest-api.yaml
```diff
@@ -2085,12 +2087,12 @@ components:
           type: integer
           description: Set to value 2 to denote a TreasuryOutput.
         amount:
-          type: integer
+          type: string
+          description: Amountt of IOTA tokens in the treasury. Plain string encoded number.
```
Thoralf-M, line 2091: Typo "Amountt"

tips/TIP-0025/core-rest-api.yaml
```diff
@@ -0,0 +1,2652 @@
...
+      properties:
+        type:
+          type: integer
+          description: Set to value 6 to denote an Immutable Alias Address Unlock Condition.
+        address:
+          oneOf:
+            - $ref: '#/components/schemas/AliasAddress'
+
+    StorageDepositReturnUnlockCondition:
+      description: Can be unlocked by depositting return amount to return address via an output that only has Address Unlock Condition.
```
grtlr, line 1804:
```suggestion
      description: Can be unlocked by depositing return amount to return address via an output that only has Address Unlock Condition.
```

Thoralf-M (commented):
Tip needs to be updated to make fields optional that are set to `0`, for example `unixTime` in an expiration unlock condition, is now skipped during the serialization if it's set to `0`

lzpap:
> Tip needs to be updated to make fields optional that are set to `0`, for example `unixTime` in an expiration unlock condition, is now skipped during the serialization if it's set to `0`

@Thoralf-M at the moment I see that none of the fields in some subschemas have the `required` keyword (for example the `Expiration Unlock Condition` schema doesn't have any required keyword . I'll have to update them. Would be sufficient to just leave such fields as they are today? (optional)

Thoralf-M:
@lzpap yes, if the other fields that are actually required will get it added, it should be fine
What also changes is the payload (currently required) and the receipts in the milestone payload (currently not even listed there)
https://github.com/iotaledger/iota.go/commit/ee47d60a05b076647b6b47e95535d5de081d278b

tips/TIP-0025/core-rest-api.yaml
```diff
@@ -0,0 +1,2701 @@
...
+              description: The current rate of referenced messages per second.
+              type: number
+              format: float
+            referencedRate:
+              description: The ratio of referenced messages in relation to new messages of the last confirmed milestone.
+              type: number
+              format: float
+        protocol:
+          description: Protocol parameters.
+          properties:
```
luca-moser, line 2408: We should extend this model with:
- `version` reflecting the protocol version
- `tokenSupply` reflecting the hex encoded token supply

as these are dynamic parameters of the protocol.

luca-moser (changes requested):

tips/TIP-0025/core-rest-api.yaml
luca-moser: Never mind, you're right, the `tokenSupply` is supposed to be a plain string, not hex encoded.

tips/TIP-0025/core-rest-api.yaml
```diff
@@ -2865,13 +2914,35 @@ components:
                 vByteFactoKey:
                   description: Defines the factor to be used for key/lookup generating fields.
                   type: integer
+            baseToken:
```
Thoralf-M, line 2917: I think this should be less indented so it's on the same level as `protocol`

tips/TIP-0025/core-rest-api.yaml
```diff
@@ -2865,13 +2914,35 @@ components:
...
+                  description: The primary unit of the token.
+                decimals:
+                  type: integer
+                  description: Number of decimals the primary unit is divisible up to.
+                subunit:
+                  type: string
+                  description: The name of the smallest possible denomination of the primary unit. subunit * 10^decimals = unit
+                "useMetricPrefix":
+                  type: boolean
+                  description: Wheteher to use metric prefixes for disaplying unit.
```
Thoralf-M, line 2937: Please also add the required fields, since `subunit` is optional

tips/TIP-0025/core-rest-api.yaml
lzpap: fixed in https://github.com/iotaledger/tips/pull/57/commits/b6cdd1e3d16ead8fdbd233b64f37aee0cc4bc151

tips/TIP-0025/core-rest-api.yaml
lzpap: fixed in https://github.com/iotaledger/tips/pull/57/commits/b6cdd1e3d16ead8fdbd233b64f37aee0cc4bc151

tips/TIP-0025/core-rest-api.yaml
```diff
@@ -0,0 +1,3356 @@
...
+            application/json:
+              schema:
+                $ref: '#/components/schemas/InternalErrorResponse'
+        '503':
+          description: "Unsuccessful operation: indicates that the node is not synced."
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ServiceUnavailableResponse'
+  '/api/v2/blocks/{blockId}/children':
```
Thoralf-M, line 324: This got removed

tips/TIP-0025/core-rest-api.yaml
```diff
@@ -0,0 +1,3352 @@
...
+        baseToken:
+          name: "IOTA"
+          tickerSymbol: "MIOTA"
+          unit: "IOTA"
+          decimals: 0
+          useMetricPrefix: true
+        features:
+          - PoW
+        plugins:
+          - indexer/v1
```
Thoralf-M, line 1222:
```suggestion
```

tips/TIP-0025/core-rest-api.yaml
```diff
@@ -0,0 +1,3354 @@
...
+                timestamp:
+                  type: integer
+                  description: The timestamp of the latest seen milestone.
+                milestoneId:
+                  type: string
+                  description: The Milestone ID of the latest seen milestone. Hex-encoded with 0x prefix.
+              required:
+                - index
+                - timestamp
+                - milestoneId
```
muXxer, line 2879:
```suggestion
```
There is an edge case where this info is not known in the node, so IMO it is better to omit these fields.
This happens for example if there is no milestone issued after bootstrapping from a genesis snapshot file, or after loading a snapshot file.

tips/TIP-0025/core-rest-api.yaml
```diff
@@ -0,0 +1,3354 @@
...
+                timestamp:
+                  type: integer
+                  description: The timestamp of the latest confirmed milestone.
+                milestoneId:
+                  type: string
+                  description: The Milestone ID of the latest confirmed milestone. Hex-encoded with 0x prefix.
+              required:
+                - index
+                - timestamp
+                - milestoneId
```
muXxer, line 2896:
```suggestion
```
There is an edge case where this info is not known in the node, so IMO it is better to omit these fields.
This happens for example if there is no milestone issued after bootstrapping from a genesis snapshot file, or after loading a snapshot file.

tips/TIP-0025/core-rest-api.yaml
```diff
@@ -0,0 +1,3333 @@
...
+          - "0xd97e89e49ed263b763afeda192515fe154ec826ceb92ef08926cd68ec02c9510"
+        isSolid: true
+        referencedByMilestoneIndex: 15465
+        whiteFlagIndex: 25
+        ledgerInclusionState: conflicting
+        conflictReason: 1
+
+    post-tagged-data-block-request-example-full:
+      value:
+        protocolVersion: 1
```
Thoralf-M, line 1331: The example should also have protocol version 2

tips/TIP-0025/core-rest-api.yaml
```diff
@@ -0,0 +1,3329 @@
...
+
+    TaggedDataPayload:
+      description: The Tagged Data Payload to be embedded into a block.
+      properties:
+        type:
+          type: integer
+          description: Set to value 5 to denote a Tagged Data Payload.
+        tag:
+          type: string
+          description: The indexation key to allow external tools to find/look up this block. It has a size between 1 and 64 bytes and must be encoded as a hex-string with 0x prefix.
```
Thoralf-M, line 2537: Should be `indexation tag` if it should be aligned with the tagFeature, but since it's not indexed by the node, maybe `indexation` should also be removed?

lzpap: fixed in https://github.com/iotaledger/tips/pull/57/commits/23a9f423480290a96c63794f356c9e29f42544e8

luca-moser (approved):

tips/TIP-0025/core-rest-api.yaml
```diff
@@ -0,0 +1,3329 @@
...
+        tag:
+          type: string
+          description: The tag to allow external tools to find/look up this block. It has a size between 1 and 64 bytes and must be encoded as a hex-string with 0x prefix. Network nodes do not index blocks with Tagged Data Payload by the tag field by default.
+        data:
+          type: string
+          description: The optional data to attach. This may have a length of 0. Hex-encoded with 0x prefix.
+      required:
+        - type
+        - tag
+        - data
```
Thoralf-M, line 2544:
```suggestion
```

tips/TIP-0025/core-rest-api.yaml
```diff
@@ -0,0 +1,3329 @@
...
+            bech32HRP:
+              type: string
+              description: Tells whether the node supports mainnet or testnet addresses. Value `iota` indicates that the node supports mainnet addresses. Value `atoi` indicates that the node supports testnet addresses.
+            tokenSupply:
+              type: string
+              description: Current supply of base token. Plain string encoded number.
+            protocolVersion:
+              type: integer
+              description: Protocol version used by the network.
+            minPoWScore:
```
grtlr, line 2936: This is not proper camelCase format (it translates to `min-Po-W-Score`). We should change it to `minPOWScore` or `minPowScore`. I prefer the later for readability.

The problem with using inconsititent formatting is that tools that automatically generate code struggle with the current format. The same thing happens in INX by the way.

tips/TIP-0025/core-rest-api.yaml
```diff
@@ -0,0 +1,3329 @@
...
+          description: The supported protocol versions.
+          items:
+            type: integer
+        protocol:
+          description: Protocol parameters.
+          properties:
+            networkName:
+              type: string
+              description: The Name of the network from which the networkId is derived.
+            bech32HRP:
```
grtlr, line 2927: Same discussion as for `minPoWScore`: https://github.com/iotaledger/tips/pull/57/files#r921955443

tips/TIP-0025/core-rest-api.yaml
```diff
@@ -0,0 +1,3327 @@
...
+              description: The primary unit of the token.
+            decimals:
+              type: integer
+              description: Number of decimals the primary unit is divisible up to.
+            subunit:
+              type: string
+              description: The name of the smallest possible denomination of the primary unit. subunit * 10^decimals = unit
+            useMetricPrefix:
+              type: boolean
+              description: Wheteher to use metric prefixes for disaplying unit.
```
Thoralf-M, line 2986:
```suggestion
              description: Whether to use metric prefixes for displaying unit.
```

tips/TIP-0025/core-rest-api.yaml
```diff
@@ -0,0 +1,3327 @@
...
+          description: The block identifiers that can be used to a attach a block to. Hex-encoded with 0x prefix.
+      required:
+        - tips
+
+    SubmitBlockRequest:
+      description: Submits a block to the node.
+      properties:
+        protocolVersion:
+          type: integer
+          description: Protocol version number of the block.It also tells which protocol rules apply to the block.
```
Thoralf-M, line 3025:
```suggestion
          description: Protocol version number of the block. It also tells which protocol rules apply to the block.
```

Thoralf-M (approved):

oopsmonk:
I think the ` Minimal Block with Tagged Data Payload` and `Minimal Block with Transaction Payload` are invalid blocks. according to TIP-00024, a block should consist of the protocol version, parents, payload, and nonce.

tips/TIP-0025/core-rest-api.yaml
lzpap: Changed in https://github.com/iotaledger/tips/pull/57/commits/103cdfa643012c0120faf4088cf5cd745e1b5de2

tips/TIP-0025/core-rest-api.yaml
lzpap: Changed in https://github.com/iotaledger/tips/pull/57/commits/103cdfa643012c0120faf4088cf5cd745e1b5de2

Thoralf-M (approved):

lzpap:
> I think the ` Minimal Block with Tagged Data Payload` and `Minimal Block with Transaction Payload` are invalid blocks. according to TIP-00024, a block should consist of the protocol version, parents, payload, and nonce.

The examples are indeed wrong. Nodes do autofill the block if some fields (parents, nonce) are missing, but protocolVersion is required and is missing in the examples. I've added them in https://github.com/iotaledger/tips/pull/57/commits/7f1e12c36f32b0fd9824a426cfcf4d26a0f0e248

tips/TIP-0025/core-rest-api.yaml
```diff
@@ -0,0 +1,3330 @@
...
+            networkName:
+              type: string
+              description: The Name of the network from which the networkId is derived.
+            bech32Hrp:
+              type: string
+              description: Tells whether the node supports mainnet or testnet addresses. Value `iota` indicates that the node supports mainnet addresses. Value `atoi` indicates that the node supports testnet addresses.
+            tokenSupply:
+              type: string
+              description: Current supply of base token. Plain string encoded number.
+            protocolVersion:
```
msarcev, line 2934: In the current node response this field is actually `version` (which imo is the better name).
The iota.js implementation follows this spec, so it won't work atm (undefined).
So we either fix iota.js and the TIP, or node api response.

lzpap: will fix this

tips/TIP-0025/core-rest-api.yaml
```diff
@@ -0,0 +1,3330 @@
...
+            params: "0x02010290394032948950303939482"
+        baseToken:
+          name: "Shimmer"
+          tickerSymbol: "SMR"
+          unit: "SMR"
+          decimals: 6
+          subunit: "glow"
+          useMetricPrefix: false
+        features:
+          - PoW
```
Thoralf-M, line 1275:
```suggestion
          - Pow
```

tips/TIP-0025/core-rest-api.yaml
```diff
@@ -0,0 +1,3330 @@
...
+              type: boolean
+              description: Whether to use metric prefixes for displaying unit.
+          required:
+            - name
+            - tickerSymbol
+            - unit
+            - decimals
+            - useMetricPrefix
+        features:
+          description: The features that are supported by the node. For example, a node could support the Proof-of-Work (PoW) feature, which would allow the PoW to be performed by the node itself.
```
Thoralf-M, line 2997:
```suggestion
          description: The features that are supported by the node. For example, a node could support the Proof-of-Work (Pow) feature, which would allow the PoW to be performed by the node itself.
```

tips/TIP-0025/core-rest-api.yaml
```diff
@@ -0,0 +1,3330 @@
...
+            protocolVersion: 2
+            params: "0x02010290394032948950303939482"
+        baseToken:
+          name: "IOTA"
+          tickerSymbol: "MIOTA"
+          unit: "IOTA"
+          decimals: 0
+          useMetricPrefix: true
+        features:
+          - PoW
```
Thoralf-M, line 1221:
```suggestion
          - Pow
```

tips/TIP-0025/core-rest-api.yaml
lzpap: @muXxer can elaborate, we decided to leave it like this for now.

PR #65, opened 2022-03-04
TIP-27 IOTA NFT standards

coodos:
This PR creates a TIP about

- IOTA NFT Metadata Standards
- ~~Decentralised PolicyID registry to prevent fake collections~~ moved to [TIP 33](https://github.com/Kami-Labs/tips/blob/tip-33/tips/TIP-0033/tip-0033.md)
- Defining Creator Royalties

[rendered version](https://github.com/Kami-Labs/tips/blob/main/tips/TIP-0027/tip-0027.md)


tips/TIP-0027/tip-0027.md
```diff
@@ -0,0 +1,291 @@
...
+- `issuerId` – UTXO string of the issuer of the NFT
+- `tokenURI` – url pointing to the NFT file location
+
+
+```json
+{
+  "IRC2, image, v1.0": {
+    // define the required utxo identifier for the NFT,
+    // the utxo identitfier for the issuer and the file location
+    "nftId": "vt7rye8tgvr7e89w",
```
lzpap, line 196: I would rather not put it here. It is already present in the output as a field.

lzpap (commented):
Thanks for the contribution, you are amazing! :rocket: 

There are some things we need to figure out but it's going in the right direction.

tips/TIP-0027/tip-0027.md
```diff
@@ -0,0 +1,320 @@
...
+  "registration": {
+    "issuerId": "s345d6f7g8h9ji07gy8h99",
+    "policyId": "a3s45dr6ft7gy87hu98ji09",
+    "policyName": "My iNFT Collection",
+    "websiteUrl": "https://mywebsite.com/policyid"
+  }
+}
+```
+
+### Creator Royalties
```
nalim2, line 114: Does it make sense to include royalties in an NFT standard? From our side, we would like to support royalties, but there are scenarios where it is necessary to have them not immutable, or it is not enough to represent them as a percentage of a sales price traded against something that is not a currency.

tips/TIP-0027/irc27.schema.json
```diff
@@ -0,0 +1,40 @@
...
+    // define the optional keys for the base schema
+    
+    // name of the asset
+    "tokenName": "My NFT #0001",
+    // unique policy ID identifier
+    "policyId": "7f9e0rwf7e90w",
+    // human readable name for the collection
+    "policyName": "My Collection of Art",
+    // Key Value pair collection of royalty addresses, where address is mapped to the percentage of payout to be sent to it
+    "royalties": {
```
adamkundrat, line 24: You should consider index here. You might want to define sequence as not all royalties might be paid every time. For example, what happens if the royalty is below dust?

tips/TIP-0027/irc27.schema.json
```diff
@@ -0,0 +1,40 @@
+{
+  // name of the standard
+  "standard" : "IRC27",
+  // type can be "image", "audio", "video", "document", "3dAsset"
+  "type": "image",
```
adamkundrat, line 5: why wouldn't you just consider standard MIME types? (https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types)

tips/TIP-0027/irc27.schema.json
```diff
@@ -0,0 +1,40 @@
...
+    "royalties": {
+      "iota1...a": 0.5,
+      "iota1...b": 0.5
+    },
+    // Name of the artist
+    "issuerName": "My Artist Name",
+    // description of the asset
+    "description": "A little information about my NFT collection",
+    // define optional attributes for NFT metadata
+    "attributes": {
```
adamkundrat, line 33: I would recommend to get inspired from ERC-721 / ERC-1155 metadata standard:
- for keys definitely should be all lower case
- you should consider support for localizations, rich formatting ( display value, css) - as recommended standard ofc

tips/TIP-0027/tip-0027.md
```diff
@@ -0,0 +1,318 @@
...
+author: Adam Eunson (@AdamCroply) <adam@tanglelabs.io>, Merul Dhiman (@coodos) <merul@tanglelabs.io>
+discussions-to: https://github.com/iotaledger/tips/discussions/59
+status: Draft
+type: Standards
+layer: IRC
+created: 2022-03-04
+---
+
+
+# iNFT Standard - IRC27
```
adamkundrat, line 14: tbh I'm not a big fun of the iNFT prefix. It wasn't even obvious to me that it's for IOTA....It's confusing. Why don't just call it NFT Standard - IRC27 ?

tips/TIP-0027/tip-0027.md
```diff
@@ -0,0 +1,318 @@
...
+
+**Universal schema**
+Each NFT schema should consist of a collection of universal keys to define key properties of the NFT
+
+The standard defined is:
+
+- `IRC27`
+
+
+The schema is defined by one of the following:
```
adamkundrat, line 171: this just feels way to limiting and I would be really careful introducing your own system for this.

adamkundrat (commented):
just add few comments! Great effort guys!

AdamCroply:
Updated a lot of the info by request and also added TIP 33 for the Token Registry

tips/TIP-0033/tip-0033.md
```diff
@@ -0,0 +1,201 @@
...
+       |__ tokens 
+       |     |__ sample-token-project-1.json
+       |     |__ sample-token-project-2.json
+       |
+       |-- NFTs
+             |__ sample-nft-project-1.json
+             |__ sample-nft-project-2.json
+```
+
+The registry will be moderated and maintained by a Committee of Human validators whose approval is needed before anything get's merged. The Committee itself will be a operated via a DAO, the DAO will control who get's to be a part of the approval committee and members can be removed or added by the DAO.
```
adamkundrat, line 76: This feels way to centralized and manual. Do we really want to have "Human validators" for this?

How do you make sure this scale, it does not hurt adoption and it's decentralized? Especially decentralization aspect needs way more details and you can't rely on just one approach. 

I understand this might be an approach by Cardano but they themselves are experiences many issues by just looking at their repo. See here: 
- https://github.com/Cardano-NFTs/policyIDs/issues/6744
- https://github.com/Cardano-NFTs/policyIDs/issues/6649
- look how many pending PR requests they have

I understand this approach and surely can work for some DAOs but it should not be enforced as primary method by IOTA (single point of record). IOTA should provide various approaches and users will decide the one they want to trust.

lzpap: I don't think the authors mean this registry as the ultimate solution, it's rather just a first step.

In the DLT world, on-chain assets are always identified by random numbers (hashes, IDs, etc.) To associate these long random numbers that humans don't ever want to see with meaningful names, there always has to be social consensus on which random number maps to what name.

At the end of the day, real world idenities must be linked to on-chain indentities. If we had all the on-chain addresses of NFT issuers published and verified off-chain, we could easily associate IRC-27 NFTs with their verified issuers, hence we can trust that the asset is legitimate. The real question is: how do you do the linking?

There are several ways:
 1. Leave it up to the user to dig up the information on the internet and decide based on the found info if they trust it or not.
 2. Have a 3rd party being responsible to do the verification off-chain, and supply the info to end users. Like twitter verified profiles.
 3. Have the community delegate experts who come together and do verification based on majority decision in an open and transparent way.
 4. Make the verification process more decentralized by having token holders vote on to-be submitted entries.
 5. Make the verification process completely automatic and decentralized by having an SC giving out verified pledges based on deterministic information. This also means that there is no social consensus, I can just buy the name `CocaColaNFT` even if I have no association whatsoever with Coca Cola.
 
All 5 are possible, but they each have their constraints. IOTA (and Shimmer) will not enforce any of these on protocol level. Applications, wallets, users should have a chance to choose which verifier service to use. TIP-27 is about what format to use to link data to an NFT, not if the data itself is legit or not.
 
[ CFNT](https://cnft.io/) is a marketplace on cardano that would fall into category `2`. There is also [Cardano Token Registry](https://github.com/cardano-foundation/cardano-token-registry) which does the same, but the trust is in the Cardano Foundation.

Based on TIP-33 my understanding is that we want to have a token registry in category `3`. Of course, it doesn't sclae well (nothing does that include social consensus), but you can trust it at least that the majority of community experts did their job with verification, there is no single point of failure.

It can also transition to `4`, or `5` with a limited set of reserved names that do require social consensus.

What do you think? How does it work in other NFT systems? Don't you have 3rd parties like opensea who verify that the issuers and the issued tokens are verified?

tips/TIP-0027/tip-0027.md
AdamCroply: The schema is expandable and versioning is implemented from the offset, introducing schema definitions is extremely advantageous for interoperability across dApps in the ecosystem

tips/TIP-0033/tip-0033.md
AdamCroply: What Levente said :D

Yes, this is not the final solution, but as an accessible primary starting point it is a viable solution and a very easily to implement, functional  deployment. In regard to the CNFT as an example, this was purely from a technical oversight, that repository is managed by a very limited dev team of just one or two validators, hence its sometimes slow pace at PR confirmations. 

With a well defined and structured decentralised committee of DAO members responsible for the validation in the IOTA ecosystem, this would be less of a concern. There are plenty of community individuals that would be more than willing to participate and support the registry. 

The main issue when you run into no.4 is this increases the delay in the verifiacation process, and with 5 how could an automated system be deployed to control IP registrations and copycat projects? If an individual is first in with $SOON token or the Soonaverse collection, it is verified and added to the registry, which I am sure wouldn't be agreeable for the Soonaverse team. With an early implementation of a decentralised committee of human validators we can negate these attack vectors with peer reviewed regitration applications.

tips/TIP-0027/tip-0027.md
AdamCroply: I love it! 💯

tips/TIP-0027/irc27.schema.json
coodos: I am not sure what you mean by localization here but if you mean glyphs from other languages than that shall be covered by unicode already

tips/TIP-0033/tip-0033.md
adamkundrat: So I’ve based on my opinion on what’s in the specification. The specification implies to the reader there will be only one system and that’s TIP-33.

It says:

> 
> To call the defined `policyId` you should request the `policyId` UTXO of the collection NFT and resolve to the Policy > Registry for human identifiable verification.
> 
> To better serve the ecosystem with a single point of record for registered collections a Public Token Registry is defined in [IOTA Public Token Registry - TIP 33](https://github.com/iotaledger/tips/pull/--) where further reading can be found.

I completely agree with above @lzpap and yes there are several systems that exists. I’m not trying to argue which one works which one not. How policy id works, etc…It might not have been very clear in my previous post. Although, I don’t think IOTA should be encouraging just one system (i.e. TIP-33) which you seem to do. It should rather focus on an open protocol and leave various policies to be implemented by third parties.

AdamCroply: There is always the potential to add future TIPs with future policies and standards. All standards and proposals are individual, there are many more standards and registries than just ERC-721 and the Ethereum Foundation Token Registry, for example. These are simply recommended first iterations, which will, in time, expand and no doubt be added further too - hence the inclusion of versioning for backwards compatibility. 

I also agree this is not the final solution, but as a starting point it makes for an easy to adopt ecosystem wide interoperable solution that can really support security, accessibility, and ease of adoption across multiple dApps from day one.

adamkundrat: That's fine. 

But again, this is not what the specification tells us. I'm very happy with you proposing TIP-33. I'm sure it'll work in some scenarios. But in my opinion it should not be mentioned in TIP-27. TIP-27 should only talk about framework that will have a support for various registries, identities and so on. It should not be pointing to just "one solution". It should not have preferred initial method. That's how it reads at the moment.

I've bounced the idea with few well established artists and they've not liked it. Have you done any analysis with the ecosystem? 

For example as a side note, Soonaverse have decided that we should not do any moderation neither try to define it as it would put is in conflict and make us rather centralized. We told community to decide on the best approaches and we will make sure to give them tools they need. We've got a specific channel on our discord (https://discord.com/channels/892481940953915433/954144769515335720). Please read this and you would notice that it's not as simple as having a moderated registry. Community is clearly looking for different solutions. 

Therefore, I believe it's safer for IF to just focus on providing a generic framework and keep TIP-33 as independent proposal that is not mixed with TIP-27.

AdamCroply: Okay, that's fine. I see your point.

lzpap (approved):
Looks good to be merged as Draft.

adamunchained (approved):
looks good, we will get this implemented within Soonaverse within next 2 - 3 weeks. We'll let you know asap if further changes are required.

huhn511:
Hey :wave: 

im thinking to remove the `collectionId` from the metadata. This could be used to scam people, if dApps not validate, if the `collectionId` is valid. So for example the NFT creator can add an `collectionId` from another NFT collection and copy it. This feature is already solved by the "issuerId" which can not be changed or add manually. Also, we need less data so we need less tokens for storage deposit. 

What do you think?

AdamCroply (approved):

Dr-Electron:
> Hey 👋
> 
> im thinking to remove the `collectionId` from the metadata. This could be used to scam people, if dApps not validate, if the `collectionId` is valid. So for example the NFT creator can add an `collectionId` from another NFT collection and copy it. This feature is already solved by the "issuerId" which can not be changed or add manually. Also, we need less data so we need less tokens for storage deposit.
> 
> What do you think?

As `collectionId` is optional I think it is ok to keep it. It can be useful sometimes for easier access to the ID but you can always also just ignore it 🤔

Dr-Electron:
I was thinking about adding something 😅 .
What about having an additional `data` property. And than you either have a uri or data property.
Because otherwise how would you use this standard if you want to store your data on-chain in the output? 🤔

adamunchained:
> I was thinking about adding something 😅 .
> What about having an additional `data` property. And than you either have to provide an uri or data property.
> Because otherwise how would you use this standard if you want to store your data on-chain in the output? 🤔

I wouldn’t complicate. You can always store it within uri. Same way as you would set src attribute in html

for example
`Red dot `

Dr-Electron:
> > I was thinking about adding something 😅 .
> > What about having an additional `data` property. And than you either have to provide an uri or data property.
> > Because otherwise how would you use this standard if you want to store your data on-chain in the output? 🤔
> 
> I wouldn’t complicate. You can always store it within uri. Same way as you would set src attribute in html
> 
> for example `Red dot`

My bad. I totally forgot the `data` URI scheme 😅

huhn511:
> As collectionId is optional I think it is ok to keep it. It can be useful sometimes for easier access to the ID but you can always also just ignore it 🤔
Yes, the `collectionId`is optional in the metadata but required as ìssuerID` on a collection. So its optional to store the data twice :P 

And i can add everything i like to the `collectionId` field - also collectionIds from other existing collections which i could use to scam people on dApps like marketplaces where this `collectionId` not gets validated. 

So why using an extra attribute which information is already present and add an potentially scam potential? Or am I misunderstanding something?

robertf26:
I agree that it does not make sense to include the `collectionId`. Besides opening the door for scammers, if it is accidentally messed up by the creator of an NFT collection and not noticed in time, there might be invalid NFT collections out there.
So why even make it complicated and give minters another possibility to make an error when we can just check the NFTid of the collection NFT?
No dapps will/should implement relying on the `collectionId` for verification if it is well known that it can be set as any arbitrary value.

Dr-Electron:
Yeah makes sense thinking about it 👍

sdellava:
An additional use case for the suggested key is a public voting process where one person can vote once. 

In these voting scenarios a voting authority is normaly requred present and required to identify the voters and authorize them to partecipate to the pool process. The vote, in most of the cases need to be secret, but this point is out of topic here.

A simple solution to provide an authorization to the voters is to mint and sent to them an NFT that have to send to a final destination address with the vote included in the tx metadata.

So in this case the allowed_address_sequence can be a list of two address: voter's address where the NFT is minted,  pool address where the NFT have to be sent.

huhn511:
Hey Stefano, thanks for your input! This sounds like an ideal use case for Smart Contracts. I see the problem, that all dApps needs to check if the address is valid or not. That means, the protocol could just send the NFT to an address which is not allowed. There is no "if/else" here on protocol level, this needs a smart contract. Because of this reason, i would say this should not be part of the NFT standard. You could extend the standard for your applications. If i miss something, please let me know. Is there anything like that on other NFT standards?

sdellava:
I understand your point. In fact, my first thought was to use a SC for these kinds of scenarios, but objectively for some of these cases it might be too much of a heavy duty.

Instead of modifying the NFT standard, it might make more sense to propose a new Unlock condition so that we can limit the displacement of a genenric token on specific addresses?

sdellava:
On second thought, both Unlock condition and Smartcontract are not safe solutions. 

In fact, Unlock conditions, if I have not misunderstood, are related to the individual tx, not to the object being moved. This implies that the next "owner" can freely define his own Unlock condition that replaces the previous one.

More or less the same applies to Smartcontract, which being deplyoed by an owner, can always be replaced.

In my initial idea these validity conditions are to be included in the immutable part of the NFT. 

Obviously it would be useful for the network to take them into account by considering invalid a tx that moves an NFT to an address not authorized by its author.

huhn511 (approved):
LGTM 👍

adamunchained:
I've spoke with @Phyloiota the other day as he asked me to review Shimmer OG NFT collection. I've provided him with feedback and he asked me to share it here so here it goes:

1. Alias should have been used to issue NFT collection. We do this because it was suggested by various people from IF after we did many reviews/testing with various parties. It's great suggestion. Alias can be used for various things (i.e. Alias can be transferred to new owner or even to a party that wants temporary do something with the collection). TIP27 does not specify how collections / NFT should be minted and I think it should be more specific!
2. Collection NFT metadata have `"type": "text/html" and "uri": "https://shimmer.network"`. This is more an additional attribute. Uri should be reserved for preview image so it can show collection banner. Right now, this NFT collection will have no image once it's imported into various NFT marketplaces.
3. I noticed Shimmer OG NFT has no props/stats. This significantly decrease value of the NFT and it's rarity. Owner will not know anything about the NFT. NFT should always have some stats/props and it should be encouraged. 
4. NFT uri could have just been `ipfs://QmPoYcVm9fx47YXNTkhpMEYSxCD3Bqh7PJYr7eo5YjLgiT` to save on storage deposit.

I'm not sure if above is something we want to enforce within the specification but it might be worth to consider some of them. Some of them can also be great "suggestions" for artists to follow.

adamunchained (changes requested):
I think we should discuss and establish specification on how NFT Collection/NFTs are minted and standardize that process.

huhn511:
Hey Adam, thanks for you feedback! 

>  TIP27 does not specify how collections / NFT should be minted and I think it should be more specific!

It's defined here - but it should be more specific and the Alias part is missing - https://github.com/Kami-Labs/tips/blob/main/tips/TIP-0027/tip-0027.md#minting-l1-nft-collections

> Collection NFT metadata have "type": "text/html" and "uri": "https://shimmer.network". This is more an additional attribute. Uri should be reserved for preview image so it can show collection banner. Right now, this NFT collection will have no image once it's imported into various NFT marketplaces.

Very Good point - so the example is a bit miss leading and we could need something like *homepage* 
``` 
"uri": "https://my-awesome-nft-project.com",
```

Dr-Electron:
> I've spoke with @Phyloiota the other day as he asked me to review Shimmer OG NFT collection. I've provided him with feedback and he asked me to share it here so here it goes:
> 
> 1. Alias should have been used to issue NFT collection. We do this because it was suggested by various people from IF after we did many reviews/testing with various parties. It's great suggestion. Alias can be used for various things (i.e. Alias can be transferred to new owner or even to a party that wants temporary do something with the collection). TIP27 does not specify how collections / NFT should be minted and I think it should be more specific!
> 2. Collection NFT metadata have `"type": "text/html" and "uri": "https://shimmer.network"`. This is more an additional attribute. Uri should be reserved for preview image so it can show collection banner. Right now, this NFT collection will have no image once it's imported into various NFT marketplaces.
> 3. I noticed Shimmer OG NFT has no props/stats. This significantly decrease value of the NFT and it's rarity. Owner will not know anything about the NFT. NFT should always have some stats/props and it should be encouraged.
> 4. NFT uri could have just been `ipfs://QmPoYcVm9fx47YXNTkhpMEYSxCD3Bqh7PJYr7eo5YjLgiT` to save on storage deposit.
> 
> I'm not sure if above is something we want to enforce within the specification but it might be worth to consider some of them. Some of them can also be great "suggestions" for artists to follow.

1. Not sure if it has any benefits in this use case. Even an NFT can be transferred. The powerful part of Alias is the unlock conditions with state and governance controller I would say.
And I think if there is a valid case to use that, it should probably be a separate TIP. This one being the easiest Implementation. In the end the NFT has some features of the Alias. So why not use it for some use-cases?
But I have no strong opinion against using an Alias. (Maybe the only one being that maybe collection NFTs are a well known thing, alias NFTs not :trollface: )
What do you mean with TIP27 doesn't specific how it should be minted? What needs to be more specific in your eyes?

2. NFTs can not just be images. So you could also specify in your guide that if someone wants to host on soonaverse he should add a image to the collection `uri` as a preview 🤔 
In the end the TIP doesn't restrict you in what to put into `uri`.
I don't think we should add a `homepage` property tho tbh. You could add such additional properties to attribute to I guess 🤔 

3. The TIP defines optional attributes. I even suggested to use them to define for example why you got the NFT (Won in Giveaway, participated in Singapur...). They didn't do it. And that's fine. we shouldn't force users to use those. But again, any marketplace could give such infos as tips in their guides. Want to increase value for your NFT? Add attributes.
After all again, not all NFTs are "just" images you try to sell.

4. This is a valid point which was actually also raised internally by @nicole-obrien afaik.

Thanks for your feedback ❤️ . I personally think most of this things should be optional.

robertf26 (approved):
LGTM, also Linter passed 👍🏼

robertf26:
Thank you to everyone who worked on TIP-27! 💯

PR #40, opened 2021-11-18
TIP-20 Transaction Payload with TIP-18 Output Types

lzpap:
## Description
[TIP-18](https://github.com/iotaledger/tips/pull/38) Introduces new output types, unlocks and output features that modify _Transaction Payload_ structure and validation rules.

This PR adds these modification to the original Transaction Payload definition (#18) with the goal of replacing [TIP-7](https://github.com/iotaledger/tips/blob/main/tips/TIP-0007/tip-0007.md).

## View

[Rendered Document](https://github.com/lzpap/protocol-rfcs/blob/tx-updates/tips/TIP-0020/tip-0020.md)

PR #69, opened 2022-03-25
TIP-29 Milestone Payload (Stardust)

Wollac:
## Description

This PR slightly modifies the original _Milestone Payload_ definition by removing the `Keys` from the `Milestone Essence` in favor of regular _Signature Blocks_ consistent with [TIP-20](https://github.com/iotaledger/tips/pull/40).

This supersedes the _Milestone Payload_ introduced in [TIP-8](https://github.com/iotaledger/tips/pull/19).

## View

[Rendered Document](https://github.com/iotaledger/tips/blob/milestone-with-signature-blocks/tips/TIP-0029/tip-0029.md)


Wollac:
Seems a bit strange it introduce a new TIP for this small change as you loose the option to view the diff. But I guess this is consistent with our process.

tips/TIP-0029/tip-0029.md
```diff
@@ -0,0 +1,242 @@
...
+            <td>uint32</td>
+            <td>The new PoW score all messages should adhere to. If 0 then the PoW score should not change.</td>
+          </tr>
+          <tr>
+            <td>Next PoW Score Milestone Index</td>
+            <td>uint32</td>
+            <td>The index of the first milestone that will require a new minimal pow score for applying transactions. This field comes into effect only if the <code>Next PoW Score</code> field is not 0.</td>
+          </tr>
+          <tr>
+            <td>Metadata</td>
```
thibault-martinez, line 116: Does this have any length requirement ? Or u16::MAX and/or max message length are default limits for it ?

Wollac: Good question. That was my initial assumption, but I checked and in [TIP-18](https://github.com/lzpap/tips/blob/master/tips/TIP-0018/tip-0018.md#metadata-block) for the _Metadata Block_ we have a `Max Metadata Length` of 8192 bytes. I don't really recall what the motivation was exactly for that limit, but then it might make sense to have this consistent.

tips/TIP-0029/tip-0029.md
```diff
@@ -0,0 +1,254 @@
...
+
+#### PoW Milestone Option
+
+The _PoW Milestone Option_ allows to change the PoW protocol parameters. 
+
+| Name                           | Type   | Description                                                                                           |
+| ------------------------------ | ------ | ----------------------------------------------------------------------------------------------------- |
+| Milestone Option Type          | uint8  | Set to **value 1** to denote a *PoW Milestone Option*.                                                |
+| Next PoW Score                 | uint32 | The new PoW score all messages should adhere to.                                                      |
+| Next PoW Score Milestone Index | uint32 | The index of the first milestone that will require a new minimal pow score for applying transactions. |
```
thibault-martinez, line 204: There used to be syntactic rules associated with this, is this not the case anymore ?

Wollac: Actually, there are even more things that need to be clarified in this context, e.g. when exactly is a message with too low PoW invalid and where is it rejected from...
Because of this, I've removed the paragraph completely it will be re-added in the upcoming Dynamic PoW TIP

lzpap (approved):
Looks good to be merged as Draft.

tips/TIP-0029/tip-0029.md
Wollac: Apparently no other options on that :disappointed: 
I would say having an implicit max (from uint16) is rather arbitrary so let's use the parameter we already have for this!

luca-moser (changes requested):
The TIP does not outline the rule that a milestone's encompassing block must have a zero nonce.

tips/TIP-0029/tip-0029.md
```diff
@@ -0,0 +1,295 @@
...
+          <tr>
+            <td>Options Count</td>
+            <td>uint8</td>
+            <td>The number of milestone options following.</td>
+          </tr>
+          <tr>
+            <td valign="top">Options <code>atMostOneOfEach</code></td>
+            <td colspan="2">
+              <details>
+                <summary>PoW Milestone Option</summary>
```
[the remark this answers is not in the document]
Wollac: Thanks

luca-moser:
Alright, then lets not add it to this TIP. Makes sense that one would have to read the block TIP anyway in order to correctly be able to construct a block with a milestone in it. @Wollac

tips/TIP-0029/tip-0029.md
```diff
@@ -0,0 +1,295 @@
...
+
+To increase the security of the design, a milestone can (optionally) be independently signed by multiple keys at once. These keys should be operated by detached signature provider services running on independent infrastructure elements. This assists in mitigating the risk of an attacker having access to all the key material necessary for forging milestones. While the Coordinator takes responsibility for forming Milestone Payload Blocks, it delegates signing in to these providers through an ad-hoc RPC connector. Mutual authentication should be enforced between the Coordinator and the signature providers: a [client-authenticated TLS handshake](https://en.wikipedia.org/wiki/Transport_Layer_Security#Client-authenticated_TLS_handshake) scheme is advisable. To increase the flexibility of the mechanism, nodes can be configured to require a quorum of valid signatures to consider a milestone as genuine.
```
muXxer, line 26: Nice to know how the coordinator is connected to its signing providers, but should that be part of the Milestone Payload TIP?

tips/TIP-0029/tip-0029.md
```diff
@@ -0,0 +1,295 @@
...
+
+### Syntactic validation
+
+Syntactic validation can be checked from the _Milestone Essence_ plus the blocks in the past cone referenced by it.
+
+- Essence:
+  - `Index Number` must not be smaller than `First Milestone Index`.
+  - If `Index Number` equals `First Milestone Index`, the following fields must be zeroed out:
+    - `Previous Milestone ID`
+    - `Inlcusion Merkle Root`
```
muXxer, line 258:
```suggestion
    - `Inclusion Merkle Root`
```

tips/TIP-0029/tip-0029.md
```diff
@@ -0,0 +1,295 @@
...
+- Forcing matching `Parents` in the _Milestone Payload_ and its block makes it impossible to reattach the same payload at different positions in the Tangle. Strictly speaking, this violates the separation of payload and block. However, it simplifies milestone processing as the position of the block will be the same as the position encoded in the Milestone Payload. Having these clear structural properties seems to be more desirable than a strict separation of layers.
+- While it is always possible to cryptographically prove that a block was confirmed by a given milestone by supplying all the blocks of a path from the milestone to the block, such a proof can become rather large (depending on the blocks). To simplify such proof-of-inclusions, the `Inlcusion Merkle Root` of all the included blocks has been added.
```
muXxer, line 291:
```suggestion
- While it is always possible to cryptographically prove that a block was confirmed by a given milestone by supplying all the blocks of a path from the milestone to the block, such a proof can become rather large (depending on the blocks). To simplify such proof-of-inclusions, the `Inclusion Merkle Root` of all the included blocks has been added.
```

tips/TIP-0029/tip-0029.md
lzpap: It doesn't hurt to have it here.

tips/TIP-0029/tip-0029.md
```diff
@@ -0,0 +1,295 @@
...
+replaces: 8
+---
+
+# Abstract
+
+In IOTA, nodes use the milestones issued by the Coordinator to reach a consensus on which transactions are confirmed. This TIP proposes a milestone payload for the blocks described in the IOTA protocol [TIP-24](../TIP-0024/tip-0024.md). It uses Edwards-curve Digital Signature Algorithm (EdDSA) to authenticate the milestones.
+
+# Motivation
+
+In order to integrate the concept of milestones consistently into Tangle blocks, this TIP describes a dedicated payload type for milestones. It contains the same essential data fields that were part of a milestone bundle in the legacy IOTA protocol. Additionally, this document also describes how Ed25519 signatures are used to assure authenticity of the issued milestones. In order to make the management and security of the used private keys easier, simple multisignature features with support for key rotation have been added.
```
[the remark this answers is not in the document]
Wollac: I've removed the reference to the legacy IOTA protocol

tips/TIP-0029/tip-0029.md
Wollac: I agree let's keep it

tips/TIP-0029/tip-0029.md
```diff
@@ -0,0 +1,295 @@
...
+
+In addition, a key rotation policy can also be enforced by limiting key validity to certain milestone intervals. Accordingly, nodes need to know which public keys are applicable for which milestone index. This can be provided by configuring a list of entries consisting of the following fields:
+- _Index Range_ providing the interval of milestone indices for which this entry is valid. The interval must not overlap with any other entry.
```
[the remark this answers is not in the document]
Wollac: I checked the Hornet config and there it seems we have (@muXxer please correct if wrong):
- A list of key ranges, each element with:
  -  public key
  - (potentially overlapping) range
-  A global config for the threshold (which is then applied to all key ranges)

Technically, it is equivalent to either have an overlapping list of single keys or a non-overlapping list of sets of key (like in the TIP).
@muXxer, I suppose you are suggesting to change the TIP to reflect the configuration format which is already implemented in Hornet?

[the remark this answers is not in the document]
muXxer: Oh, I misunderstood that part. Thats even better than it is implemented right now in hornet, because you can specify the threshold per range. And it's easier to read.

So the TIP is fine, maybe we should change the format in hornet in an upcoming version.

muXxer (approved):

PR #93, opened 2022-08-30
Enforce lowercase on node features

muXxer:
There were several discussions about the casing of `PoW` in the info endpoint.

It is safer to enforce lowercase for every feature on the node side for easier string comparison on the client side.

Hint: The client side could also convert all features it receives via the info endpoint to lowercase before comparing with the needed feature, to make the check more fail safe against non-conform node implementations.

### DIFF
--- a/tips/TIP-0025/core-rest-api.yaml
+++ b/tips/TIP-0025/core-rest-api.yaml
@@ -1218,7 +1218,7 @@ components:
           decimals: 0
           useMetricPrefix: true
         features:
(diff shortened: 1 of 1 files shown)


lzpap (approved):

thibault-martinez (approved):
I am not fully convinced we should care about non-conform node implementations by lowercasing from the client but looks good anyway.

alexsporn (approved):

PR #101, opened 2022-09-13
Create Generic Signed Payload

sdellava:
As quickly discussed, I'm happy to submit this very simple TIP that will allow to build a standard Selective Permanode INX plug-in

### DIFF
--- a/Generic Signed Payload
+++ b/Generic Signed Payload
@@ -1,0 +1,62 @@
+---
+tip: <to be assigned>
+title: <Generic Signed Paload>
+description: <Standard format of a Tangle message containing a payload signed by the publisher>
+author: <Stefano Della Valle (@sdellava)>
+discussions-to: <https://github.com/iotaledger/tips/discussions/100>
+status: Draft
+type: <Standard>
+layer (*only required for Standards Track): <Core>
+created: <2022-9-13>
+requires (*optional): <TIP number(s)>
+replaces (*optional): <TIP number(s)>
+superseded-by (*optional): <TIP number(s)>
+---
+
+## Abstract
+Many applications that generate and publish data can benefit from the IOTA network as a transport layer. The STREAM protocol/library is designed to fulfill this purpose, ensuring that the sender is identified and thus that messages containing false data cannot be created.
+However, for many low-cost applications STREAM can be too expensive and complex while writing data into the generic payload of a Tange Message (TIP-6) is very simple.
+
+In addition the pruning of the tangle performed by the nodes make the message reception unreliable.
+
+The solution is to create a simple INX plug-in for the node that can filter messages received by the node based on their tag and store them in a local database and then provide them to the user on demand. 
+
+Obviously, the user will have to worry about deleting superfluous messages from this database and providing the node with sufficient storage space. 
+To confirm that the messages that are stored are correct and real, the node also calculates and stores a Proof of Inclusion for each message.
+
+This approach can also be used for other purposes, however, it is subject to a simple attack: the tag, the only element that can be used to distinguish messages without getting into the paylod, can be arbirally compiled, so a malicious publisher can create fake messages that the receiver cannot recognize.
+
+
+## Motivation
+
+To support low-cost applications we need to create a selective permanode inx plug-in, and to make it reliable we need a standard method for identifying fake messages.
+
+This TIP is intended to define a message format that can be recognized as true or false by a standard method.
+
+
+## Specification
+
+Detailed design
+
+The following describes the standard method for creating a generic message containing data whose validity can be verified by the receiver.
+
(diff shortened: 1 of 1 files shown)


lzpap (commented):
Hey, @sdellava, couple remarks:
 - `messages` are called `blocks`, as per [TIP-24].(https://github.com/iotaledger/tips/blob/main/tips/TIP-0024/tip-0024.md)
 - Since this is an application level standard, it should not introduce a new payload type into the core protocol. You can use the [Tagged Data Payload](https://github.com/iotaledger/tips/blob/main/tips/TIP-0023/tip-0023.md) to store/publish your application specific data structure.
 - Change the `layer` of the TIP to IRC, see the different layers in in [TIP-1](https://github.com/iotaledger/tips/blob/main/tips/TIP-0001/tip-0001.md#tip-types).
 - Define the structure, layout and representation of the data you'd like to put into the `Data` field of the Tagged Payload.
 - Explain the interface and components of the INX application you are planning to build.
 - Explain how the INX application filters blocks, what syntactic/semantic checks should be implemented.
 - Inlcude a rough API design on how the INX application exposes data through an API.
 - Inlcude example payloads and their validation.

lzpap (commented):
Also, please update the PR folder structure to match the [TIP process requirements](https://github.com/iotaledger/tips/blob/main/tips/TIP-0001/tip-0001.md#tip-workflow).

sdellava:
Thanks for your feedback. 

In an initial evaluation I assumed that the public key could be used instead of the tag, however, it is evident that the tag allows different types of data contents produced by the same author to be specified. Therefore, I propose to create a new payload type (which I have tentatively denoted 6) in which the signature and public key are present in addition to the tag.

alexsporn:
There is no real difference in computation overhead for a L2 solution to extract the signature from a protocol field or from the payload itself. I see no reason to pollute the L1 protocol with fields with signatures that it will not verify.

Explain to me the difference between this 2 usecases:

1. Current Tagged Data:
 * Tag: `myspecialpayload`
 * Data: `mydatapayload+signature(mydatapayload)`
 
 2. The proposed solution
 * Tag: `myspecialpayload`
 * Data: `mydatapayload`
 * Signature: `signature(mydatapayload)`
 
If the argument is, that `Tag` cannot be spammed with solution 2, then this is wrong, because I can also send invalid signatures and L2 will need to verify them anyway to filter these out.

sdellava:
Defining a Layer 2 format that must use the Tag defined in the Layer 1 format to discriminate the meaning of the Layer 2 content is in my opinion much more wrong than creating a specific Layer 1.

True, there is no particular difference between the two options, but this is a motivation that reinforces this approach. 

If nodes ignore the contents of blocks with payload type 5, they also ignore those with this payload type, so there is no side effect for the network. 

Having this type of payload also makes it possible to create particular, very lightweight nodes that might instead ignore all other types of blocks and consider only those that have the special nature of not being retained by "normal" nodes.

Possibly, as an alternative and compromise, we can simply modify the Tagged Payload, adding after the date field, the key and signature as optional fields. In this way, a new payload format is not created but we reinforce payload type 5 by making it more robust and versatile.

alexsporn:
> Possibly, as an alternative and compromise, we can simply modify the Tagged Payload, adding after the date field, the key and signature as optional fields. In this way, a new payload format is not created but we reinforce payload type 5 by making it more robust and versatile.

Adding fixed optional fields does not make it more versatile. You have a clean canvas inside the `Data` payload where you can specify any fields for any use-case. Adding specific fields for special use-cases just pollutes the L1.

Currently you are thinking about 2 extra fields, _public key_ and _signature_. But what happens then when the next requirement arrives and it wants to add a UUID to the data? Or a validity timerange? Or some reference to another previous block?

All this is application logic, not L1 logic, and thus should be handled on the application level.

If you want to standardise payloads for certain use-cases, to be interoperable with other applications, then that is what the IRC standards are for. Just look hat the current proposed NFT standards, it is exactly the same idea.

sdellava:
The problem I see with creating a standard L2 format is that you basically need to define a also L2 payload tyoe to allow the receiver to understand how to interpret the data contained in the L2 payload. In example, the singed data playload is type 0 and involves key and signature, while type 1 contains an encrypted message and thus contains key, signature, and public key of the recipient.

In other words, since there are no restrictions the content of an L2 payload is ambiguous.

To resolve this ambiguity the only option I see is to make the use of the typed L2 payload mandatory for the L1 payload type 5. However, this requirement is equivalent to creating a new L1 payload type.

sdellava:
To be clear and not to lose the focus of the problem: if we cannot be certain of the reliability of a Tagged Block, we cannot store it in a selective permanode.

The meaning of the payload is irrelevant to its preservation. The only thing that matters is that it is authentic, and the only way to verify this is to allow the sender to sign it.

PR #115, opened 2022-11-22
Fix: Add syntactic validation rule for max data length

brancoder:
- Added Tagged Data Payload syntactic validation rule for max length of the data field.
- Updated Syntactic rules to reference protocol parameters for Iota and Shimmer.


brancoder:
Not needed for the syntactic validation rule for max length of the data field as the length of the data field is bounded by the block size.

PR #116, opened 2022-12-07
Fix sender and issuer field name

Thoralf-M:


EddyTheCo:
Related to this. 
The 'StorageDepositReturnUnlockCondition' json object specified in tip-25 do not agrees with the current client and node expected value.
The tip-25 expect a 'returnAmount' string while the nodes and the client expects 'amount'.

If you want to change the tips to agree with the node and client implementation, I will only change the tips describing the expected Json object(tip-25, tip26) and not change the tips describing the serialization of the object like tip-18, tip-20.

lzpap (approved):

muXxer (approved):

PR #117, opened 2023-01-01
Microtransactions on layer 1 do not need the sender feature

EddyTheCo:
To unlock a storage deposit unlock condition, the unlocking transaction has to deposits to   the `Return Address` value not to the `sender ` feature value. This commit fix that.


EddyTheCo:
What happen when the return amount(from the Storage Deposit Return Unlock Condition) is less that the min_deposit_of_output of the simplest output?

Can Alice create an output such that the amount of the output is greater than the min_deposit_of_output of that output, but the return amount is lees that min_deposit_of_output of a basic output with only an address unlock condition? 
If the later is possible Bob will not be able to consume the output? 

By reading the tips  it seems this is possible.  But if Bob can not consume that output there is not point to allow this on the protocol.

lzpap (approved):
Thanks for the patch @EddyTheCo, this was left there from earlier iterations.

> What happen when the return amount(from the Storage Deposit Return Unlock Condition) is less that the min_deposit_of_output of the simplest output?

As for your question, it is not possible to create such outputs due to the [syntactic validation rules of the `Storage Deposit Return Unlock Condition`.](https://github.com/iotaledger/tips/blob/main/tips/TIP-0018/tip-0018.md#additional-syntactic-transaction-validation-rule)

PR #119, opened 2023-01-09
RestAPI changes to TIP-0013 and TIP-0025

muXxer:
We added new endpoints to `/included-message` and `/included-block` for easier usage.
Basically these are the same endpoints that exist for `/messages/{msgID}/` or `/blocks/{blockID}/` as well, but for the message/block that got included for a specific `transactionId`

### DIFF
--- a/tips/TIP-0013/rest-api.yaml
+++ b/tips/TIP-0013/rest-api.yaml
@@ -464,10 +464,10 @@ paths:
               schema:
                 $ref: '#/components/schemas/OutputResponse'
               examples:
-                Output Unspent:
-                  $ref: '#/components/examples/get-outputs-by-id-response-example'
-                Output Spent:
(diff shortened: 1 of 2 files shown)


tips/TIP-0013/rest-api.yaml
```diff
@@ -2448,6 +2618,12 @@ components:
             isSpent:
               type: boolean
               description: Tells if the output is spent or not.
+            milestoneIndexSpent:
+              type: integer
+              description: The milestone index at which this output was spent.
+            transactionIdSpent:
+              type: string
+              description: The transaction this output was spent with. Hex-encoded with 0x prefix.
```
alexsporn, line 2626: We don't have the 0x encoding on chrysalis

alexsporn (changes requested):

lzpap (approved):

PR #133, opened 2023-06-20
feat: add TIP-48: Metadata Encoding Standard v1

Tuditi:
A standard for encoding and decoding data in the metadata field of an output.


PhilippGackstatter:
Just two notes: The TIP number conflicts with #153 (and we already have links to the IOTA 2.0 TIP-48 in many places, so it would probably be much easier to update the MES-TIP number) and the metadata in IOTA 2.0 will be a [map](https://github.com/iotaledger/tips/blob/tip38/tips/TIP-0038/tip-0038.md#metadata-feature) rather than a flat array, so you could consider adapting the TIP to that structure.

On the other hand, one _possible_ use of the metadata map is so that each application-level use case can use its own key and define its encoding/compression/format/... in its own spec or TIP. We could consider having a "Registry TIP" where applications that want to interop can register their key. This would not be protocol enforced, of course, but these kinds of registries seem to work pretty well (a large scale example being [SLIP-44: Coin Types](https://github.com/satoshilabs/slips/blob/master/slip-0044.md)).
For instance, as far as I'm aware, it is planned that `did:iota`, which is currently stored directly in the state metadata of an alias output, migrates to one of the keys.

PR #134, opened 2023-07-21
Fix Unlock Syntactic Validation

Thoralf-M:
A reference index must be < `Max Inputs Count - 1`, because it otherwise could be a self reference, which is not allowed in the unlocks semantic validation. So no point in allowing that in the syntactic validation of an unlock then in the first place


Thoralf-M:
@lzpap there it's already correct https://github.com/iotaledger/iota.go/blob/d9965e47e73db490acca2c1ede03d66965e9c1db/utxo_input.go#L15

Thoralf-M:
The link above was for Chrysalis, actually let's keep it as is and change only in 2.0

PR #140, opened 2023-10-23
TIP-38: Building Blocks for IOTA 2.0 Output Types #22 

PhilippGackstatter:
[Rendered Version](https://github.com/iotaledger/tips/blob/tip38/tips/TIP-0038/tip-0038.md)

Note that links to other new TIPs don't work and will only work once the TIP is merged.


tips/TIP-0038/tip-0038.md
```diff
@@ -0,0 +1,928 @@
...
+
+## Addresses
+
+The following section shows the basic supported Address Types and their serialization.
+
+| Address | Type Byte as `uint8` | Bech32 Encoded |
+| ------- | -------------------- | -------------- |
+| Ed25519 | 0                    | iota1**q**...  |
+| Account | 8                    | iota1**p**...  |
+| NFT     | 16                   | iota1**z**...  |
```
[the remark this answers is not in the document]
PhilippGackstatter: I generally avoid any kind of backlinking references because it's a nightmare to maintain.
If another TIP comes around in the future that adds another address type, this list will once again be outdated and hardly anyone will remember to update it.
So I generally try to write TIPs so that they can remain stable and don't need to be updated / become outdated even when new TIPs are proposed, with the trade-off that we never have a global list of all address types or all features types, ... in one TIP.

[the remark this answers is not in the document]
Thoralf-M: Not sure if I follow, this table was defined in TIP 18 already
In the TIP it mentions `Many of these were originally introduced in [TIP-18](https://github.com/iotaledger/tips/blob/tip38/tips/TIP-0018/tip-0018.md), and the functionality defined in this document is an extension of the primitives defined in TIP-18.`
Shouldn't there then rather be a table with only the two new addresses?
Also in https://github.com/iotaledger/tips/blob/tip38/tips/TIP-0038/tip-0038.md#address-unlock-condition all addresses are listet

I see the additional part for the table is here for the restricted address https://github.com/iotaledger/tips-draft/blob/tip50/tips/TIP-0050/tip-0050.md#restricted-address

but in https://github.com/iotaledger/tips-draft/blob/tip52/tips/TIP-0052/tip-0052.md#bech32-representation there is nothing (at the bottom an example where one can see it's `y`, but not as this table form)

On a side note, why has TIP 50 the higher (40) type byte than TIP52 (32)? Don't care that much about it, just noticed it now :trollface: 

Don't get me wrong, I don't have a big issue with anything here, just some observation
So feel free to resolve the conversation if you think it's good as it is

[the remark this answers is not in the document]
PhilippGackstatter: Well, TIP-18 is completely deprecated for IOTA 2.0 so we have to redefine this stuff, and for the those three address types that happens in this TIP.

Nice observation about the missing bech32 table, I just added it to TIP-52! 

When we started the TIPs my goal was also to have them build on each other successively without and circular references, but that is not possible due to many small details. Some unlock conditions need to reference output types (like Storage Deposit Return UC needs to define that the return amount must be deposited via a Basic Output), so already we have a TIP-38 -> TIP-41 dependency. And there are many more of those, unfortunately.

So that led me to drop the efforts of trying to increase type numbers in line with the TIPs. Also, we try for them to make sense on the type range they're defined on, which is not necessarily the order of the TIPs. One can only optimize for one logical dimension here :smile:.

tips/TIP-0038/tip-0038.md
```diff
@@ -0,0 +1,1064 @@
...
+the receiver address could technically issue a backdated block shortly after `ATT - Liveness Threshold Upper Bound`, but
+this block would be less likely to be selected as a tip, and because the majority of online validators already accepted
+the expired output consumption, they should not accept the backdated receiver consumption because it is a conflict. The
+only way the receiver address could revert the other conflict is by having support of a majority of other validators to
+revert it. In conclusion, it is very difficult to backdate even with just acceptance, and becomes more difficult when
+waiting for higher levels of finality. This attack relates to the fundamental fact that the best approximation of time
+is the block's `Issuing Time` timestamp, which is is meant to be very difficult to falsify.
+
+**Note**: An IOTA Smart Contract Chain is not susceptible to such a backdating attack anyway because it is the receiver
+address of the expiration locked output and it is _not_ waiting for the output to expire in order to consume it.
```
Thoralf-M, line 1060: But ISC could also create outputs with an expiration unlock condition?

PhilippGackstatter: Do they actually do that?

And even if they did, when they are the `Return Address` and wait for the expired output to be committed (for example), they have a reasonably high guarantee that they can consume it without the `Address` owner backdating a block and consuming it.

Thoralf-M: Currently not that I'm aware of, but it could be possible
Just don't understand why there is an extra note for this

PhilippGackstatter: We discussed it in the context of ISC, but it's indeed a bit out of place here, so I removed it.

tips/TIP-0038/tip-0038.md
```diff
@@ -0,0 +1,1062 @@
...
+      </details>
+      <details>
+        <summary>Restricted Address</summary>
+        <blockquote>An address that contains another address and allows for configuring its capabilities. Defined in <a href='../TIP-0050/tip-0050.md#restricted-address'>TIP-50 (Restricted Address)</a>.</blockquote>
+      </details>
+    </td>
+  </tr>
+</table>
+
+_Restricted Addresses_ cannot be put into _Sender Features_ because they are resolved to the underlying address during
```
thibault-martinez, line 716: Then don't put it in the list?

tips/TIP-0038/tip-0038.md
```diff
@@ -0,0 +1,1054 @@
...
+Commitment the Return Address owner can pick is 20-3 = 17, and because 20 <= 17+3 is true, they can unlock.
+
+If however, the current slot is 18, and both owners would use the same Commitment Input with index 13, within the range
+of 8 and 15 of possible commitments, then neither one of them can unlock the output, as neither 20 > 13+10 nor 20 <=
+13+3 is true.
+
+**Backdating blocks**
+
+One variable that was not mentioned yet is how a client can set a block's `Issuing Time`, relative to which the chosen
+commitment is validated, which allows for _backdating_ blocks. Backdating blocks by a singificant amount is only
```
thibault-martinez, line 1024:
```suggestion
commitment is validated, which allows for _backdating_ blocks. Backdating blocks by a significant amount is only
```

tips/TIP-0038/tip-0038.md
```diff
@@ -0,0 +1,1054 @@
...
+      <b>Description</b>
+    </td>
+  </tr>
+  <tr>
+    <td>Unlock Condition Type</td>
+    <td>uint8</td>
+    <td>Set to <strong>value 0</strong> to denote an <i>Address Unlock Condition</i>.</td>
+  </tr>
+  <tr>
+    <td valign="top">Address <code>oneOf</code></td>
```
thibault-martinez, line 370: It is not very clear if something is omitted because it's contextual (like ImplicitAccountCreationAddress only being allowed in Basic outputs) or because it simply is disallowed (like I assume ImplicitAccountCreationAddress is in StorageDepositReturn and Expiration?).

thibault-martinez: So maybe put an explicit note in StorageDepositReturn and Expiration that ImplicitAccountCreationAddress is disallowed or in TIP 42 be a bit more explicit that it can only be used in an AddressUnlockCondition.

thibault-martinez: Ignore me if it's said already somewhere, may have missed it.

tips/TIP-0038/tip-0038.md
```diff
@@ -0,0 +1,1054 @@
...
+    outputs of the native tokens determines whether the tokens are burned (removed from the ledger) or melted in the
+    foundry.
+    - The transaction is invalid if Native Tokens are burned and the _Can Burn Native Tokens_ flag in the _Transaction
+      Capabilities_ is **unset**. Melting is unaffected by the flag.
+
+# Migration
+
+## Native Token Migration
+
+The capability to hold Native Tokens is removed from Alias and NFT Outputs. In Basic and Foundry Outputs the capabaility
```
thibault-martinez, line 967:
```suggestion
The capability to hold Native Tokens is removed from Alias and NFT Outputs. In Basic and Foundry Outputs the capability
```

tips/TIP-0038/tip-0038.md
PhilippGackstatter: It's tricky to express this correctly here where Address UC is generically defined, because it suggest that this always is the schema of an Address UC, which is a reasonable assumption given that it's true for every other schema.

Every output currently expresses which address is allowed in its Address UC (Basic allows Impliciti Acc Creation Addresses while others don't).

- One option is to omit the addresses here and adding a note that each output specifies which addresses it allows. 
- Another one is to only include the ones that are always allowed to be present and add a note that each occurence of the Address UC can specify additional ones or something along those lines. Basically how it is now, minus the note. I think this one I like more because it doesn't break the schema definitions as much as the other one.

thibault-martinez: Ok but if we do that, it means that I have to look at ALL tips to see if something is allowed or not. Like, how do I come to the conclusion whether ImplicitAddress is allowed in a StorageDeposit or Expiration? If it just doesn't make sense to start with, I would put a note in this TIP.

PhilippGackstatter: I'd say you can come to the conclusion that Implicit Account Creation Address it not allowed in the SDRUC because it's not in the SDRUC schema and there's **no note** that occurences of the SDRUC may change that.
Conversely, for Address UC I would add such a note.

I know it's somewhat implicit that way, but it's also not sensible to add a note to every schema that no changes are allowed to it. So in essence, I'd make the changes opt-in by adding a note (like for Addres UC).

Does that sound reasonable?

tips/TIP-0038/tip-0038.md
```diff
@@ -0,0 +1,1071 @@
...
+            <td>(uint16)ByteArray</td>
+            <td>An array of arbitrary binary data. A leading uint16 denotes its length.</td>
+          </tr>
+        </table>
+      </details>
+    </td>
+  </tr>
+</table>
+
+#### Additional Transaction Syntactic Validation Rules
```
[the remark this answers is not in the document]
PhilippGackstatter: Yes, I think so, since e.g. Transaction and Block IDs depend on the serialization. Hence, all implementations should produce the same serialization of a Transaction and hence features.

tips/TIP-0038/tip-0038.md
```diff
@@ -0,0 +1,1086 @@
...
+not require fees. To prevent the receiving party from blocking access to the storage deposit, it is advised to be used
+together with the [Expiration Unlock Conditions](#expiration-unlock-condition). The receiving party then has a
+sender-defined time window to agree to the transfer by consuming the output, or the sender regains total control after
+expiration.
+
+### Timelock Unlock Condition
+
+Slot indices in the Tangle are introduced via slot commitments and represent a notion of time. Each such commitment
+carries an index. When using any feature related to time, a
+[_Commitment Input_](../TIP-0045/tip-0045.md#commitment-input) is needed as a reference of time.
```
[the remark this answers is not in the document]
PhilippGackstatter: Yes, only then.

PR #143, opened 2023-10-23
TIP-41: Basic Output Type

PhilippGackstatter:
[Rendered Version](https://github.com/iotaledger/tips/blob/tip41/tips/TIP-0041/tip-0041.md)

Note that:


tips/TIP-0041/tip-0041.md
```diff
@@ -0,0 +1,575 @@
...
+
+### Created Outputs
+
+- All <i>Unlock Condition</i> imposed transaction validation criteria must be fulfilled.
+- All <i>Feature</i> imposed transaction validation criteria must be fulfilled.
+
+# Test Vectors
+
+The protocol parameters used in the following test vectors are the same as in
+[TIP-49 (Protocol Parameters Hash)](../TIP-0049/tip-0049.md#protocol-parameters-hash).
```
DaughterOfMars, line 519: dead link. probably the same in all of these examples

tips/TIP-0041/tip-0041.md
```diff
@@ -0,0 +1,575 @@
...
+
+## Storage Score
+
+The following test vector shows the calculation of the storage score according to [TIP-47](../TIP-0047/tip-0047.md).
```
DaughterOfMars, line 523: also dead

tips/TIP-0041/tip-0041.md
thibault-martinez: Because the PR is not merged yet, it will work later

DaughterOfMars: oh 🤦‍♀️

PhilippGackstatter: Yeah, sorry about that. I think I might have to put in temporary static github links for CI to pass before merging anyway, so I might update those when I have spare time.

PR #146, opened 2023-10-23
TIP-44: Foundry Output Type

PhilippGackstatter:
[Rendered Version](https://github.com/iotaledger/tips/blob/tip44/tips/TIP-0044/tip-0044.md)

Note that:


tips/TIP-0044/tip-0044.md
```diff
@@ -0,0 +1,610 @@
...
+
+The following test vector shows the calculation of the storage score according to [TIP-47](../TIP-0047/tip-0047.md).
+
+Foundry Output (json-encoded):
+
+```json
+{
+  "type": 3,
+  "amount": "420000000",
+  "serialNumber": 0,
```
DaughterOfMars, line 568: Serial number cannot be zero

PhilippGackstatter: Updated, thanks!

PR #147, opened 2023-10-23
TIP-45: IOTA 2.0 Transaction Payload

PhilippGackstatter:
[Rendered Version](https://github.com/iotaledger/tips/blob/tip45/tips/TIP-0045/tip-0045.md)

Note that:


tips/TIP-0045/tip-0045.md
```diff
@@ -0,0 +1,1101 @@
...
+[semantic validity](#semantic-validation) of the encapsulating _Transaction Payload_ is unaffected by the possibly
+contained payload.
+
+### Unlocks
+
+The `Unlocks` field holds the unlocks which unlock the inputs in a _Transaction_.
+
+#### Signature Unlock
+
+The <i>Signature Unlock</i> holds a signature signing the BLAKE2b-256 hash of the <i>Transaction</i> (including the
```
thibault-martinez, line 653: Can you elaborate what is being signed exactly? Is it just the hash of the transaction serialization bytes or is it one of the TransactionId components?

PhilippGackstatter: Clarified in the TIP. It's the serialized `Transaction` field.

tips/TIP-0045/tip-0045.md
```diff
@@ -0,0 +1,1611 @@
...
+  - Outputs:
+    - `Outputs Count` must be `0 < x ≤ Max Outputs Count`.
+    - For each output the following must be true:
+      - `Output Type` must match one of the values described under [Outputs](#outputs).
+      - The `Amount` field must not be `0`.
+      - The `Amount` field must be at least the minimum storage deposit according to [TIP-47](../TIP-0047/tip-0047.md).
+      - The output itself must pass syntactic validation.
+    - The sum of all `Amount` fields must not exceed `Max IOTA Supply`.
+    - The sum of all `Mana` fields must not exceed <code>2<sup>Mana Bits Count</sup> - 1</code>.
+    - The count of all distinct native tokens present in outputs must not be larger than `Max Native Token Count`.
```
thibault-martinez, line 1199: I think we said we would remove this limitation now since it's one token per output we're just bound to the number of ouput per tx

PhilippGackstatter: Yeah and I can't find that check anymore so I think it was removed :+1:.

tips/TIP-0045/tip-0045.md
```diff
@@ -0,0 +1,1610 @@
...
+
+- Let `Size Score` be `Work Score Parameters::DataByte * Transaction Size` where `Transaction Size` is the size of the
+  serialized _Transaction_.
+- Let `Input Score` be `Inputs Count * Work Score Parameters::Input`.
+- Let `Context Input Score` be `Context Inputs Count * Work Score Parameters::Context Input`.
+- Let `Output Score` be the sum of work scores of each output `Output` in `Outputs`:
+  - Let `Native Token Score` be `Work Score Parameters::Native Token * Native Tokens Count` if `Output` contains
+    `Native Tokens`, `0` otherwise.
+  - Let `Token Scheme Score` be `Work Score Parameters::Native Token` if `Output` contains a _Simple Token Scheme_, `0`
+    otherwise.
```
Alex6323, line 1145: Is that still the case?

tips/TIP-0045/tip-0045.md
```diff
@@ -0,0 +1,1610 @@
...
+
+Let the work score of a transaction be defined as follows.
+
+- Let `Size Score` be `Work Score Parameters::DataByte * Transaction Size` where `Transaction Size` is the size of the
+  serialized _Transaction_.
+- Let `Input Score` be `Inputs Count * Work Score Parameters::Input`.
+- Let `Context Input Score` be `Context Inputs Count * Work Score Parameters::Context Input`.
+- Let `Output Score` be the sum of work scores of each output `Output` in `Outputs`:
+  - Let `Native Token Score` be `Work Score Parameters::Native Token * Native Tokens Count` if `Output` contains
+    `Native Tokens`, `0` otherwise.
```
Alex6323, line 1143: needs to be updated I guess.

tips/TIP-0045/tip-0045.md
```diff
@@ -0,0 +1,1610 @@
...
+  - Let `Token Scheme Score` be `Work Score Parameters::Native Token` if `Output` contains a _Simple Token Scheme_, `0`
+    otherwise.
+  - Let `Staking Score` be `Work Score Parameters::Staking` if `Output` contains a _Staking Feature_, `0` otherwise.
+  - Let `Block Issuer Score` be `Work Score Parameters::Block Issuer` if `Output` contains a _Block Issuer Feature_, `0`
+    otherwise.
+  - Return
+    `Work Score Parameters::Output + Native Token Score + Token Scheme Score + Staking Score + Block Issuer Score`.
```
Alex6323, line 1150: just to make sure: adding the `Output` offset from the `WorkScoreParameters` is still needed for each output, right?

tips/TIP-0045/tip-0045.md
PhilippGackstatter: Yes :)

tips/TIP-0045/tip-0045.md
PhilippGackstatter: Looking into this now, we're somehow not using the Output offset anymore in the code which is odd.

Alex6323: Is there any update on this?

PhilippGackstatter: We are most likely missing the parameter in our calculations and the intention is to have it, so the TIP is correct, but the node currently isn't. Tracked in https://github.com/iotaledger/iota-core/issues/564.

tips/TIP-0045/tip-0045.md
```diff
@@ -0,0 +1,1605 @@
...
+
+Syntactic validation is checked as soon as the transaction has been received. It validates the structure but not the
+signatures of the transaction. If the transaction does not pass this stage, it must not be broadcast further and can be
+discarded right away.
+
+The following criteria defines whether a payload passes the syntactical validation:
+
+- Transaction:
+  - `Network ID` must match the value of the current network.
+  - `CreationTime` must be a valid `Slot index` timestamp.
```
thibault-martinez, line 1178: Needs to be changed to creation slot

thibault-martinez: Just checking in, you resolved but didn't remove?

tips/TIP-0045/tip-0045.md
```diff
@@ -0,0 +1,1646 @@
...
+    <td>The commitment identifier to reference to.</td>
+  </tr>
+</table>
+
+#### Block Issuance Credit Input
+
+A _Block Issuance Credit Input_ provides the BIC balance of a specific account as context to transaction execution of a
+specific slot. It is required for any _Account_ transition and destruction if the account contains a _Block Issuer
+Feature_, as any operation on the _Account_ can only be allowed if the BIC balance is non-negative at a specific point
+in time. A _Block Issuance Credit Input_ always requires a _Commitment Input_. The input will be resolved to the BIC
```
thibault-martinez, line 870: > A _Block Issuance Credit Input_ always requires a _Commitment Input_.
This doesn't seem to be true, the node accepts BIC input without a commitment input.

tips/TIP-0045/tip-0045.md
```diff
@@ -0,0 +1,1646 @@
...
+
+- Transaction:
+  - `Network ID` must match the value of the current network.
+  - Inputs:
+    - `Inputs Count` must be `0 < x ≤ Max Inputs Count`.
+    - For each input the following must be true:
+      - `Input Type` must denote an allowed input type, listed in section [Inputs](#inputs).
+      - `Transaction Output Index` must be `0 ≤ x < Max Outputs Count`.
+    - Each pair of `Transaction ID` and `Transaction Output Index` must be unique in the list of inputs.
+  - Context Inputs:
```
thibault-martinez, line 1184: Should these be sorted?

PhilippGackstatter: Since the insertion order of arrays is retained, we don't absolutely need these to be sorted. On the other hand, lexical ordering makes duplicate checking more (memory) efficient, so it seems beneficial to add it. I would do the same for `Inputs` then.

tips/TIP-0045/tip-0045.md
PhilippGackstatter: Do you happen to have an example or a alphanet transaction that shows that?
I don't see how it could get past this check: https://github.com/iotaledger/iota-core/blob/de711606b447c36733ee60d3b72a0932166d07c1/pkg/protocol/engine/ledger/ledger/vm.go#L96.

tips/TIP-0045/tip-0045.md
```diff
@@ -0,0 +1,1753 @@
...
+it does not have it.
+
+| Flag Index | Capability (if flag is set)  |
+| ---------- | ---------------------------- |
+| 0          | Can burn Native Tokens.      |
+| 1          | Can burn Mana.               |
+| 2          | Can destroy Account Outputs. |
+| 3          | Can destroy Anchor Outputs.  |
+| 4          | Can destroy Foundry Outputs. |
+| 5          | Can destroy NFT Outputs.     |
```
thibault-martinez, line 994: I guess not but just making sure, so we need a flag for Delegation burning?

PhilippGackstatter: No flag needed. The Reward Input basically acts like the flag. If it's missing, the tx fails. And since destoying a Delegation Output is a more typical operation than destroying other chain outputs, we don't add an extra flag.

thibault-martinez: What I thought, thanks 👍🏻

PR #152, opened 2023-10-24
TIP-52: Multi Address

PhilippGackstatter:
[Rendered Version](https://github.com/iotaledger/tips/blob/tip52/tips/TIP-0052/tip-0052.md)

Note that:


tips/TIP-0052/tip-0052.md
```diff
@@ -0,0 +1,341 @@
...
+
+# Multi Address
+
+The following table shows the mapping from the address type of the **first byte** to the address type:
+
+| Address       | Type Byte as `uint8` | Bech32 Encoded |
+| ------------- | -------------------- | -------------- |
+| Multi Address | 40                   | iota1**q**...  |
+
+The following table shows the serialization of a _Restricted Address_:
```
thibault-martinez, line 53:
```suggestion
The following table shows the serialization of a _Multi Address_:
```

tips/TIP-0052/tip-0052.md
```diff
@@ -0,0 +1,341 @@
...
+          </tr>
+          <tr>
+            <td valign="top">Address <code>oneOf</code></td>
+            <td colspan="2">
+              <details>
+                <summary>Ed25519 Address</summary>
+                <blockquote>An Address derived from an Ed25519 Public Key. Defined in <a href='../TIP-0038/tip-0038.md#ed25519-address'>TIP-38 (Ed25519 Address)</a>.</blockquote>
+              </details>
+              <details>
+                <summary>Account Address</summary>
```
thibault-martinez, line 107: What about Anchor ?
And Restricted ?

tips/TIP-0052/tip-0052.md
```diff
@@ -0,0 +1,345 @@
...
+A transaction containing a Multi Unlock is only valid if all of the following conditions hold:
+
+- The corresponding Address of the to-be-unlocked input is of type _Multi Address_.
+- The `Addresses Count` of the Multi Address and the `Unlocks Count` of the Multi Unlock must match.
+- If `Unlocks` contains Reference, Account or NFT Unlocks, they must point to an Unlock outside of the Multi Unlock.
+- Let the `Cumulative Unlocked Weight` be the sum of weights of the addresses that were successfully unlocked, through
+  Unlocks other than _Empty Unlocks_, which do not add to this sum.
+- `Cumulative Unlocked Weight >= Threshold`.
+
+### Empty Unlock
```
thibault-martinez, line 227: Should it be a syntactic rule that it can't be used outside of a MultiUnlock or are we fine waiting semantic and see that address/unlock is not a good match?

PhilippGackstatter: The schemas tell you where the Empty Unlock is allowed - only in Multi Unlocks.
Same for implicit account creation addresses. They are also not allowed in Address Unlock Conditions within Accounts, but are allowed in Basic Outputs, neither are they allowed in Expiration Unlock Conditions. Those rules are not defined anywhere, the schema itself is the rule.

Since we already have the schemas and they are much easier to maintain in the TIPs I would like to avoid duplicating those rules in text were it cannot be automatically updated.

tips/TIP-0052/tip-0052.md
```diff
@@ -0,0 +1,345 @@
...
+- Modelling a 1 of n control structure in addition to a single backup address that has full control.
+- Allowing interested parties to inspect the conditions for unlocking, e.g. majority or supermajority.
+
+# Multi Address
+
+The following table shows the mapping from the address type of the **first byte** to the address type:
+
+| Address       | Type Byte as `uint8` | Bech32 Encoded |
+| ------------- | -------------------- | -------------- |
+| Multi Address | 40                   | iota1**q**...  |
```
thibault-martinez, line 51: The example `iota19qq0ezu97zl76wqnpdxxleuf55gk0eqhscjtdgqm5sqwav6gcarz6vvesnk` given in this TIP doesn't adhere to this?

PhilippGackstatter: Byte 40 is encoded to `9` in bech32, I might've confused that for a `q` :sweat_smile:. Updated.

tips/TIP-0052/tip-0052.md
```diff
@@ -0,0 +1,345 @@
...
+</table>
+
+A _Multi Address_ is a container for other addresses. Each address has weight and the Multi Address contains a threshold
+`Threshold` that needs to be reached by the unlocked addresses in order to unlock the Multi Address.
+
+## Additional syntactic transaction validation rules
+
+A Multi Address is only valid if all of the following conditions hold:
+
+- `1 <= Addresses Count <= 10`.
```
thibault-martinez, line 143:
```suggestion
- `2 <= Addresses Count <= 10`.
```
https://github.com/iotaledger/iota.go/pull/647

PR #156, opened 2023-11-02
TIP-53: Tagged Data

PhilippGackstatter:
[Rendered Version](https://github.com/iotaledger/tips/blob/tip53/tips/TIP-0053/tip-0053.md)

Note that:

### DIFF
--- a/tips/TIP-0053/tip-0053.md
+++ b/tips/TIP-0053/tip-0053.md
@@ -1,0 +1,76 @@
+---
+tip: 53
+title: Tagged Data Payload
+description: Block payload for arbitrary data
+author: Philipp Gackstatter (@PhilippGackstatter) <philipp.gackstatter@iota.org>
+discussions-to: https://github.com/iotaledger/tips/pull/156
+status: Draft
+type: Standards
+layer: Core
+created: 2023-11-02
+replaces: 23
+---
+
+# Abstract
+
+This TIP is a slightly revised version of TIP-23 with updates to IOTA 2.0, like the the type prefix of the _Tagged Data_ Payload type and its syntactic validation.
+The payload concept offers a very flexible way to combine and encapsulate information in the IOTA protocol. This document proposes a basic payload type that allows the addition of arbitrary data.
+
+# Motivation
+
+The most flexible way to extend an existing object is by the addition of arbitrary data. This payload provides a way to do just that. An optional tag can be used to categorize the data.
+
+# Specification
+
+## Tagged Data
+
+The following table describes the serialization of a _Tagged Data Payload_ following the notation from [TIP-21](../TIP-0021/tip-0021.md):
+
+<details>
+  <summary>Tagged Data</summary>
+  <blockquote>Data with an optional tag.</blockquote>
+</details>
+<table>
+  <tr>
+    <td>
+      <b>Name</b>
+    </td>
+    <td>
+      <b>Type</b>
+    </td>
+    <td>
+      <b>Description</b>
+    </td>
+  </tr>
+  <tr>
+    <td>Payload Type</td>
+    <td>uint8</td>
+    <td>Set to <strong>value 0</strong> to denote a <i>Tagged Data Payload</i>.</td>
+  </tr>
+  <tr>
+    <td>Tag</td>
+    <td>(uint8)ByteArray</td>
+    <td>The tag of the data.</td>
+  </tr>
+  <tr>
+    <td>Data</td>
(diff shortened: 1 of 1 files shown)


tips/TIP-0053/tip-0053.md
```diff
@@ -0,0 +1,84 @@
...
+author: Philipp Gackstatter (@PhilippGackstatter) <philipp.gackstatter@iota.org>
+discussions-to: https://github.com/iotaledger/tips/pull/156
+status: Draft
+type: Standards
+layer: Core
+created: 2023-11-02
+replaces: 23
+---
+
+# Abstract
```
thibault-martinez, line 14:
```suggestion
# Summary
```
You've used Summary for all the other TIPs

tips/TIP-0053/tip-0053.md
```diff
@@ -0,0 +1,84 @@
...
+  </tr>
+</table>
+
+It is important to note that `Tag` is not considered by the protocol, it just serves as a marker for second layer
+applications.
+
+## Syntactic Validation
+
+- The length of the `Tag` must not be larger than `64`.
+- The length of `Data` must not be larger than `8192`.
```
thibault-martinez, line 72: Just to make sure, both tag and data can have length 0, right ? because the description says `Data with an optional tag` which makes it seem data can't be empty.

tips/TIP-0053/tip-0053.md
```diff
@@ -0,0 +1,84 @@
...
+- Given the type and length information, the _Tagged Data Payload_ must consume the entire byte array of the `Payload`
+  field of the encapsulating object.
```
thibault-martinez, line 74: Kinda comes back to the same payload length discussion but tbh this rule doesn't make a lot of sense to me. At least not in this TIP IMO. There is nothing to check in the context of the payload itself.

PhilippGackstatter: Yeah, I think this should be a general serialization rule. After deserialization no bytes should remain. TIP-21 would be the place to write that down, but I don't want to modify or replace that one.

In any case, I'll remove it here.

tips/TIP-0053/tip-0053.md
PhilippGackstatter: I checked with the others and it's indeed optional. Updated the TIP to clarify.

tips/TIP-0053/tip-0053.md
```diff
@@ -0,0 +1,82 @@
...
+# Specification
+
+## Tagged Data
+
+The following table describes the serialization of a _Tagged Data Payload_ following the notation from
+[TIP-21](../TIP-0021/tip-0021.md):
+
+<details>
+  <summary>Tagged Data</summary>
+  <blockquote>Optional Data with an optional tag.</blockquote>
```
thibault-martinez, line 35:
```suggestion
  <blockquote>Optional data with an optional tag.</blockquote>
```

tips/TIP-0053/tip-0053.md
```diff
@@ -0,0 +1,82 @@
...
+status: Draft
+type: Standards
+layer: Core
+created: 2023-11-02
+replaces: 23
+---
+
+# Summary
+
+This TIP is a slightly revised version of TIP-23 with updates to IOTA 2.0, like the the type prefix of the _Tagged Data_
```
thibault-martinez, line 16:
```suggestion
This TIP is a slightly revised version of TIP-23 with updates to IOTA 2.0, like the type prefix of the _Tagged Data_
```

thibault-martinez (approved):

tips/TIP-0053/tip-0053.md
```diff
@@ -0,0 +1,82 @@
...
+    <td>Binary data.</td>
+  </tr>
+</table>
+
+It is important to note that `Tag` is not considered by the protocol, it just serves as a marker for second layer
+applications.
+
+## Syntactic Validation
+
+- The length `Len` of the `Tag` field must satisfy: `0 <= Len <= 64`.
```
Alex6323, line 71: out of interest: why this tag length limitation?

PhilippGackstatter: I think it has historical reasons. I believe it was limited to `64` when tags were still indexed by the protocol (Chrysalis), presumably due to optimizing for a certain database implementation and it has simply carried over through Stardust and IOTA 2.0. We could allow a larger tag length as far as I'm aware, but no one has really asked for it.

PR #157, opened 2023-11-02
TIP-54: Anchor Output Type

PhilippGackstatter:
[Rendered Version](https://github.com/iotaledger/tips/blob/tip54/tips/TIP-0054/tip-0054.md)

Note that:

### DIFF
--- a/tips/TIP-0054/tip-0054.md
+++ b/tips/TIP-0054/tip-0054.md
@@ -1,0 +1,652 @@
+---
+tip: 54
+title: Anchor Output Type
+description: Defines the IOTA 2.0 Anchor Output used to anchor commitments into Layer 1.
+author:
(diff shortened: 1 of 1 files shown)


tips/TIP-0054/tip-0054.md
```diff
@@ -0,0 +1,868 @@
...
+      "type": 5,
+      "address": {
+        "type": 8,
+        "accountId": "0x17432c5a7a672503480241125e3952414a7a320441080c624c264b004e09614a"
+      }
+    }
+  ],
+  "features": [
+    {
+      "type": 0,
```
DaughterOfMars, line 834: Features can only contain metadata

PhilippGackstatter: Updated, thanks for catching that.

